Issue #11, opened 2022-12-20
 Write a script called build.sh that will create the thyra-plugin-wallet Binary, taking in two arguments: the port number and the path to the plugin.


B-Naoufal:

SlnPons:
@B-Naoufal can you add more details about what's the purpose of this issue? 
To be able to install the plug-in easily, is that correct? Cheers

B-Naoufal:
@SlnPons the goal is to have a build file build.sh, which can be called by the CI in this repository 
thyra-plugin-wallet and produces in input a binary (executable file ), thyra will after automatically download this binary file, and install the plugin wallet:  

Any Pull Request in [thyra-plugin-wallet](https://github.com/massalabs/thyra-plugin-wallet) ==> Merge ==> CI ==> call build.sh ==> Produce thyra-plugin-wallet Binary.  
So When you launch thyra:  
Call Plugin Manager ==> Download the last Produced thyra-plugin-wallet Binary By thyra-plugin-wallet CI

SlnPons:
Hey @B-Naoufal I failed at accessing the plug-in wallet as described above. 
Here is what I did: 
1. Download the binary on this page: https://github.com/massalabs/thyra-plugin-wallet/actions/runs/3764513478
2. Execute the binary using the following cmd on my terminal `wallet-plugin chmod +x thyra-plugin-wallet`
3. Try to access the wallet plugin on port 8080 as described on my terminal (see screenshot below)


![Image](https://user-images.githubusercontent.com/109611779/209657922-c6317dfa-c795-4e04-91a3-7e869152fd1b.png)


4. Get a 404 error

PR #12, opened 2022-12-20
Fix CI/CD and build the compatible binary connecting wallet as plug-in to thyra

B-Naoufal:
1- Write a script called build.sh that will create the thyra-plugin-wallet Binary, taking in two arguments: the port number and the path to the plugin.  
2- Fix The CD calling the build.sh, and upload artifacts for multiple Operating Systems.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,7 +43,7 @@ jobs:
           submodules: true
       - uses: actions/setup-go@v3
         with:
-          go-version: ">=1.19.2"
+          go-version: ">=1.15"
       - name: Build binary and prepare archive for publish
         shell: bash
         run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,6 @@ permissions:
   contents: read
 
 jobs:
-
   lint:
     uses: ./.github/workflows/lint.yml
 
@@ -20,7 +19,6 @@ jobs:
     uses: ./.github/workflows/tests.yml
 
   build:
-    if: ${{ github.ref == 'refs/heads/main' }}
     needs: [lint, tests]
     name: build and upload artifacts
     strategy:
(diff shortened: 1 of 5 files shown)


build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -ex
+# Set the GOPATH environment variable
+PLUGIN=wallet-plugin
+
+
+# Build the binary
+mkdir -p "build/$PLUGIN/"
+env CGO_ENABLED="1" go build -o build/$PLUGIN/thyra-plugin-wallet  cmd/massa-wallet/main.go
```
peterjah, line 9: what is "env". i think you can remove it

.github/workflows/build.yml
```diff
@@ -1,71 +1,86 @@
...
+    # The lint job uses a configuration file in the .github/workflows directory.
     uses: ./.github/workflows/lint.yml
 
   tests:
+    # The tests job uses a configuration file in the .github/workflows directory.
     uses: ./.github/workflows/tests.yml
 
   build:
-    if: ${{ github.ref == 'refs/heads/main' }}
+    # The build job depends on the lint and tests jobs to run successfully before it can start.
```
peterjah, line 28: please add this:
    if: ${{ (github.ref == 'refs/heads/main') || startsWith(github.ref, 'refs/tags/v') }}

We want to build binaries only on main branch and releases

peterjah (changes requested):

.github/workflows/build.yml
B-Naoufal: ok but we may have some broken CI in main after merge :? how to test that the build is working well in all targets before merge ? @peterjah

B-Naoufal: yes you're right

peterjah (approved):

build.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -ex
```
gregLibert, line 2:
```suggestion
#!/bin/bash -ex
```

cmd/massa-wallet/main.go
```diff
@@ -14,14 +15,19 @@ import (
 )
 
 func main() {
-	myApp := app.New()
+	var port int
+	var path string
+	flag.IntVar(&port, "port", 8080, "the port to listen on")
+	flag.StringVar(&path, "path", "", "the path to listen on")
```
gregLibert, line 21: Why ?

gregLibert (changes requested):
:mag: nitpicking and unused variable

gregLibert (approved):
:clap:

SlnPons:
@B-Naoufal same question as for #11 not sure how to test it correctly.

PR #30, opened 2023-01-18
29 have a wallet encoding iso with massa client

SebastienValla:


pkg/wallet/wallet.go
```diff
@@ -21,14 +21,16 @@ const (
 	PBKDF2NbRound             = 10000
 	FileModeUserReadWriteOnly = 0o600
 	Base58Version             = 0x00
+	SaltSize                  = 12
```
gregLibert, line 24: Why? According to NIST, the salt must be at least 128 bits long. By reducing it to 96 bits, you reduce the number of attempts needed to test all values by 2^36, or 68,719,476,736.

Source to do the math: Why? According to NIST, the salt must be at least 128 bits long. By reducing it to 96 bits, you reduce the number of attempts needed to test all values by 2^36, or 68,719,476,736.

gregLibert (changes requested):
:no_good: security reasons.

pkg/wallet/wallet.go
SebastienValla: so there is a safety issue in massa-client, I will contact the core team
https://github.com/massalabs/massa/blob/main/massa-cipher/src/constants.rs

AurelienFT: Our salt can't be more than 64 because it's the limit for the crate we use (we can change crate). Here is their explanations on why they placed 64 as a limit : https://docs.rs/password-hash/latest/password_hash/struct.Salt.html#constraints

gregLibert: Sounds good. We just need it to be 16 bytes long (128 bits). It's what we did on our side :-)

AurelienFT: Ok I misread "bits" and "bytes". We can change it to 16 on our side too

SebastienValla:
waiting for core team to move on to unblock this issue

SlnPons:
I think this PR will be unblocked once https://github.com/massalabs/massa-standards/issues/13 is done. 
@SebastienValla @Thykof so you agree?

Thykof:
> I think this PR will be unblocked once [massalabs/massa-standards#13](https://github.com/massalabs/massa-standards/issues/13) is done. @SebastienValla @Thykof so you agree?

No, once https://github.com/massalabs/massa-standards/issues/15 is done

Issue #31, opened 2023-02-08
api not working on the "function rest wallet list"


noesusset:
**Describe the bug**

The "rest wallet list" api calls doesnt return the informations about the KeyPair such as "nonce", "privateKey"...

**To Reproduce**

after installing the thyra-plugin-wallet repo run
```
go run cmd/massa-wallet/main.go
```

then we can create a wallet 
and using for example postman we can call the api and see that there is some missing information

![image](https://user-images.githubusercontent.com/96011618/217648854-d39ba9bf-4d68-41ac-908e-60d482f1a1a9.png)

**Expected behavior**

we should be able to see the information about the KeyPair

SlnPons:
@Thykof @gregLibert is it fair to say that this issue will be solved / a proper solution will be defined with the wallet standard? 

If not, what can we do so solve this issue on the wallet plugin please?

gregLibert:
> @Thykof @gregLibert is it fair to say that this issue will be solved / a proper solution will be defined with the wallet standard?

Nope

> If not, what can we do so solve this issue on the wallet plugin please?

This issue shouldn't be solved. the wallet plugin shouldn't extract key pair. if the caller need to sign something, the Sign endpoint should be used instead.

PR #32, opened 2023-02-08
fix: fixed the list of wallets request

noesusset:
fixed the api calls such that it can now return the KeyPair values

image

this PR should fix this issue  #31

### DIFF
--- a/internal/handler/wallet/get.go
+++ b/internal/handler/wallet/get.go
@@ -1,6 +1,7 @@
 package wallet
 
 import (
+	"github.com/btcsuite/btcutil/base58"
 	"github.com/go-openapi/runtime/middleware"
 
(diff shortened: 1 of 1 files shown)


noesusset:
#32

SlnPons:
Hello @noesusset, sorry for the delay. I forgot to check my github notifications this month... 😅
And thank you very much for collaborating but your PR doesn't go in the direction we would like to take for the wallet plugin. Indeed, we want to provide the highest security possible for our users. For that, no key pairs should be shared like this but rather thyra would be the intermediary between you and the wallet; as thyra is installed on your machine there will be nothing between you and the blockchain and so the private key that you want to keep private will stay private no matter what! 

In other terms, the wallet plugin won't get the private key nor share it. Hope it's clear, feel free to ping me if not :) 
Cheers

PR #37, opened 2023-02-23
Migrate to new plugin management

Thykof:


web/html/wallet.js
```diff
@@ -6,6 +6,10 @@ closeModalOnClickOn("close-button");
 closeModalOnClickOn("nicknameCancelBtn");
 getWallets();
 
+function addPrefixUrl(relativeURL) {
+    return `http://my.massa/thyra/plugin/Massalabs/Massa%20Wallet/${relativeURL}`
```
Thykof, line 10: this part smells bad I would welcome your improvement suggestions

README.md
```diff
@@ -12,10 +14,38 @@ When you start this plugin, which you can do using the command `go run cmd/massa
...
+```shell
+./build.sh
+```
+
+Install the plugin:
+
+```shell
+mkdir ~/.config/thyra/my_plugins/thyra-plugin-wallet
+PLUGIN=wallet-plugin
+mv build/$PLUGIN/thyra-plugin-wallet ~/.config/thyra/my_plugins/thyra-plugin-wallet
```
gregLibert, line 50: I'm not sure that you need to install the plugin list that :-)

README.md
```diff
@@ -12,10 +14,38 @@ When you start this plugin, which you can do using the command `go run cmd/massa
...
+Run:
+
+```shell
+go run cmd/massa-wallet/thyra-plugin-wallet.go 1 --standalone
+```
+
+Build:
+
+```shell
+./build.sh
```
gregLibert, line 42: This file is not needed anymore I guess

cmd/massa-wallet/thyra-plugin-wallet.go
```diff
@@ -44,11 +35,35 @@ func StartServer(app *fyne.App, port int) {
...
+	//nolint:gomnd
+	if len(os.Args) < 2 {
+		panic("this program must be run with correlation id argument!")
+	}
+
+	pluginID := os.Args[1]
+
+	standaloneMode := false
+
+	if len(os.Args) == 3 && os.Args[2] == "--standalone" {
```
gregLibert, line 59: This flag is not describe. Furthermore I think that it should be an env variable instead.

cmd/massa-wallet/thyra-plugin-wallet.go
```diff
@@ -44,11 +35,35 @@ func StartServer(app *fyne.App, port int) {
...
+	}
+	registerPlugin(ln)
 
 	if err := server.Serve(); err != nil {
-
 		log.Fatalln(err)
 	}
 }
+
+func registerPlugin(ln net.Listener) {
```
gregLibert, line 49: This code is the very same for all module.
It should be define at the hello-world level and a function should be imported here to avoid copy/paste

cmd/massa-wallet/thyra-plugin-wallet.go
```diff
@@ -44,11 +35,35 @@ func StartServer(app *fyne.App, port int) {
...
+	//nolint:gomnd
+	if len(os.Args) < 2 {
+		panic("this program must be run with correlation id argument!")
```
gregLibert, line 52: You should use an "Usage: ..." message instead .

cmd/massa-wallet/thyra-plugin-wallet.go
```diff
@@ -44,11 +35,35 @@ func StartServer(app *fyne.App, port int) {
...
+	}
+
+	pluginID := os.Args[1]
```
gregLibert, line 55: Not sure it makes sense to ask the enduser to define a plugin id when using standalone mode

pkg/plugin/register.go
```diff
@@ -0,0 +1,78 @@
+package plugin
```
gregLibert, line 1: SHould be imported from hello-world :-)

web/html/wallet.js
gregLibert: Can't you generate that from the register function ? My point is that you are defining the author and plugin name twice...

gregLibert (changes requested):
:no_good: We should reuse things here to avoid to much duplication.

README.md
Thykof: this is the manual way of installing the plugin in the "contribut" section, only for development purpose

CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
...
+We welcome and appreciate contributions from the community. Whether it's a bug report, feature request,
+code contribution, or anything else, we're grateful for your effort. Here's a brief guide to get you started.
+
+Any contribution you make will be reflected on [CONTRIBUTORS](CONTRIBUTORS.md).
+
+To get an overview of the project, read the [README](README.md).
+
+## Product Vision
+
+[Project Name] is a [description of the project]. Our goal is to [outline the project's goal].
```
Thykof, line 13: @SlnPons I would need you for this section ;)

cmd/massa-wallet/thyra-plugin-wallet.go
Thykof: you right, I have twisted hello world a bit here https://github.com/massalabs/thyra-plugin-hello-world/pull/8

CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
...
+- Submit design improvements
+
+## Code of Conduct
+
+Help us keep this project open and inclusive and keep our community approachable and respectable.
+
+## Bug Reports
+
+If you find any bugs, feel free to open an issue in the
+[issue tracker](https://github.com/[user/org]/[project-name]/issues). Please include as much information as possible,
```
gregLibert, line 31: Url to fix :-)

CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
...
+such as operating system, browser version, etc.
+
+## Feature Requests
+
+If you have an idea for a new feature, feel free to open an issue in the
+[issue tracker](https://github.com/[user/org]/[project-name]/issues). Please include as much information as possible,
```
gregLibert, line 37: same

CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
...
+such as what the feature would do, why it would be useful, etc.
+
+## Code Contributions
+
+We are always looking for developers to help us improve the codebase. If you are interested in contributing code, please
+take a look at the [open issues](https://github.com/[user/org]/[project-name]/issues) and follow the steps below:
```
gregLibert, line 43: same

CONTRIBUTING.md
```diff
@@ -0,0 +1,64 @@
...
+- Write clean, easy-to-understand code.
+- Write tests for your code in order to ensure that it works as expected.
+- Make sure all tests pass before submitting a pull request.
+- Follow our code conventions and formatting standards.
+- Document your code thoroughly.
+
+## Questions
+
+If you have any questions, feel free to open an issue in the
+[issue tracker](https://github.com/[user/org]/[project-name]/issues).
```
gregLibert, line 62: same

web/html/constants.js
```diff
@@ -0,0 +1,2 @@
+// auto generated file by gen.go
```
gregLibert, line 1: Should it be commited if it's go generated?

gregLibert: Why do you need this file ? Why don't you:
- define an endpoint to return these constant at the helloworld level
- bind a new endpoint in this project using the handler of hello world ?

gregLibert (changes requested):

web/html/constants.js
Thykof: yes it must be commited to follow good practices.

I have generated this file from golang code so that the constants are only defined at one place: golang code, and accessible directly in the javascript code. (btw I know that the front end will be entirely rework)

I don't understand you proposal: because the constants are deffirent from one plugin to another, why the wallet would use a handler of the hello-world? Also, my front end app need these constants to perform any http call to thyra so I can't get these variable through API call.

CONTRIBUTING.md
SlnPons: @Thykof done, just pushed my suggestion :)

gregLibert (approved):

gregLibert:
:clap:

mazmassa:
LGTM! 👍

mazmassa (approved):

PR #39, opened 2023-03-01
update address format

SebastienValla:


pkg/wallet/wallet.go
```diff
@@ -244,7 +244,7 @@ func CreateWalletFromKeys(nickname string, privateKey []byte, publicKey []byte,
 	wallet := Wallet{
 		Version:  0,
 		Nickname: nickname,
-		Address:  "A" + base58.CheckEncode(addr[:], Base58Version),
+		Address:  "AU" + base58.CheckEncode(addr[:], Base58Version),
```
gregLibert, line 247: You need to do the same thing for the function Import, line 218

gregLibert (changes requested):
:no_good: one update is missing

gregLibert (approved):

ASAPSegfault (approved):

PR #48, opened 2023-03-13
allow CORS

Thykof:


mazmassa (approved):
LGTM!

ASAPSegfault (approved):

cmd/massa-wallet/thyra-plugin-wallet.go
```diff
@@ -34,6 +35,10 @@ func startServer(app *fyne.App) {
 	server := restapi.NewServer(massaWalletAPI)
 	server.ConfigureAPI()
 
+	if os.Getenv("STANDALONE") == "1" {
```
mazmassa, line 38: Non blocking: isn't better have this names mapped in some global constants (if it exists in Go) 🤔 !?

```const STANDALONE = "STANDALONE"```

... and then in the code we could use this global constant instead of the string value it self. This would allow us have only one source of truth for possible future changes instead of replace each occurrence of "STANDALONE".

Feel free to resolve!

mazmassa (approved):
LTGT

cmd/massa-wallet/thyra-plugin-wallet.go
Thykof: good catch thanks! I'll do it in the next PR.

gregLibert: Two things:
- in Go global constant are only possible for static/compilation time value. here you will have to use a cosntant I guess and having a global constant is a bad practice (in all language). But you could do the same thing using a function returning this value
- I think this is the only place where the value is used, so I guess it's maybe overkill.

mazmassa: > ...But you could do the same thing using a function returning this value or a boolean

Ahh, ok! Thank you! Good to know!

> * I think this is the only place where the value is used, so I guess it's maybe overkill.

Indeed!

Thykof: @gregLibert I already implemented here https://github.com/massalabs/thyra-plugin-hello-world/pull/10/files

SlnPons:
@0xMazout could you tell me if it's working now please? 
If so, just put this issue in done :D cheers

0xMazout:
We can launch massa tube and see if we have an error in console related to cors it's not but i think it's working if we analyze the code . let's test it when we have time

PR #61, opened 2023-03-27
move tests to related folder

mazmassa:
We are moving the tests files to a specific related folder in order to keep a clean structure.

<table><tr><td><b>Before</b></td> <td><b>After</b></td></tr><tr>
<td> 

```js
\wallet
    - file1.go
    - file1_test.go
    - file2.go
    - file2_test.go
    - ...  
```

</td>
<td>

```js
\wallet
    - file1.go
    - file2.go
    - ...
    \tests
        - file1_test.go
        - file2_test.go
        - ...  
```

</td>
</tr>
</table>


gregLibert:
> We are moving the tests files to a specific related folder in order to keep a clean structure.
> 
> Before	After
> ```js
> \wallet
>     - file1.go
>     - file2.go
>     - ...
>     \tests
>         - file1_test.go
>         - file2_test.go
>         - ...  
> ```
> 
> ```js
> \wallet
>     - file1.go
>     - file1_test.go
>     - file2.go
>     - file2_test.go
>     - ...  
> ```

Two things:
- your code is actually doing the opposite of what you said
- go is expecting to have both test and code file in the same directory. Changing that is maybe possible but will be certainly painful at some point.

gregLibert (changes requested):
:no_good:

mazmassa:
> * your code is actually doing the opposite of what you said
I did no code. Just moved to a /tests folder.

> * go is expecting to have both test and code file in the same directory. Changing that is maybe possible but will be certainly painful at some point.
Really!? We can't have that in a specific folder!?

mazmassa:
> 🙅

Haven't got the point here neither!? That means this should not be done!? It's wrong!? Can I have more info!?

mazmassa:
Unfortunately that's the way it works on GO! Same dir for tests... :( 

Closing it!

PR #65, opened 2023-03-29
Fix build

Thykof:


.github/workflows/build.yml
```diff
@@ -2,6 +2,7 @@
 on:
   push:
     branches: [ main ]
+  pull_request:
```
mazmassa, line 5: This is a mystery for me... do we really need it!? 🤔

mazmassa (approved):
LGTM!

.github/workflows/build.yml
Thykof: yes because I want to now if the build work before merging into main branch (this is what broke currently)

mazmassa: Got it! I see! Thank you! 🙏 

Let me understand better then... shouldn't we specify the types of the PRs to be more selective? 🤔 

```
pull_request:
   types: [opened, reopened, synchronize, ready_for_review, labeled, review_requested, edited]
```

 Omitting that is same to validate for all types? Is that the aim!?

0xMazout (approved):
Look like good , i didn't tested it 🦸 
If it's working LGTM

PR #70, opened 2023-04-04
Implement account file format standard

Thykof:


pkg/wallet/wallet.go
```diff
@@ -18,21 +17,20 @@ import (
...
+	"gopkg.in/yaml.v2"
 	"lukechampine.com/blake3"
 )
 
 // TODO: rename to 'account'
 
 const (
 	SecretKeyLength           = 32
-	PBKDF2NbRound             = 600000
+	PBKDF2NbRound             = 60_0000
```
thomas-senechal, line 28: I guess there is an issue here
```suggestion
	PBKDF2NbRound             = 600_000
```

Thykof: good catch

PR #76, opened 2023-04-05
use mnemonic

peterjah:


pkg/wallet/wallet.go
```diff
@@ -371,7 +391,10 @@ func Import(nickname string, privateKeyB58V string, password string) (*Wallet, e
 	return wallet, nil
 }
 
-func CreateWalletFromKeys(nickname string, privateKey []byte, publicKey []byte, password string) (*Wallet, error) {
+func CreateWalletFromKeys(nickname string, privateKey ed25519.PrivateKey, password string) (*Wallet, error) {
+
+	publicKey := reflect.ValueOf(privateKey.Public()).Bytes() // force conversion to byte array
```
Thykof, line 396: great! with privateKey.Public() of package ed25519 we can use ed25519 algorithm (or curve) to create our public key :)

pkg/wallet/wallet.go
```diff
@@ -286,16 +288,27 @@ func loadFile(filePath string) (Wallet, error) {
...
+	}
+
+	mnemonic, err := bip39.NewMnemonic(entropy)
 	if err != nil {
-		return nil, fmt.Errorf("generating ed25519 keypair: %w", err)
+		return nil, fmt.Errorf("creating mnemonic: %w", err)
 	}
+	fmt.Println("mnemonic")
+	fmt.Println(mnemonic)
+	privateKey := bip39.NewSeed(mnemonic, "")
```
Thykof, line 303: error here: bip39.NewSeed returns a 64 bytes private key, we want 32 bytes

maybe we can take the 32 first:

```golang
privateKey := bip39.NewSeed(mnemonic, "")[:32]
```

pkg/wallet/wallet.go
```diff
@@ -335,10 +348,19 @@ func FilePath(nickname string) (string, error) {
 	return path.Join(configDir, Filename(nickname)), nil
 }
 
+func ImportFromMnemonic(nickname string, mnemonic string, password string) (*Wallet, error) {
```
Thykof, line 351: this function is not used in the code yet (only in unit test)

pkg/wallet/wallet.go
Thykof, line 303: it is discouraged to directly use seed as a private key

```suggestion
	seed := bip39.NewSeed(mnemonic, "")
	privateKey := sha512.Sum512(seed)
```

PR #87, opened 2023-04-11
Change location of account files

Thykof:


pkg/wallet/wallet.go
```diff
@@ -213,23 +214,35 @@ func GetConfigDir() (string, error) {
...
 
+// GetWalletDir returns the path to the wallet directory.
+// Note: the wallet directory is the folder where the wallet plugin binary resides.
+func GetWalletDir() (string, error) {
+	configDir, err := GetConfigDir()
+	if err != nil {
+		return "", fmt.Errorf("reading config directory '%s': %w", configDir, err)
+	}
+
+	return path.Join(configDir, plugin.PluginDirectoryName, "wallet-plugin"), nil
```
peterjah, line 225: I am not very comfortable with this. Here we assume that plugin use the Thyra config dir. but imho, the plugin should be independant from Thyra.
What i would do is to wirte everything in the same folder as te binary, wherever it is. so it will work with thyra and in standalone mode.
It will allow to remove dependencies from thyra (constant names).
It will also allow us to use the wallet plugin in standalone mode without using the same files as the thyra wallet plugin installed.

peterjah (changes requested):

pkg/wallet/wallet.go
Thykof: nice catch

0xMazout (approved):

peterjah (approved):

PR #95, opened 2023-04-13
Add CORS for DELETE

Thykof:


api/server/restapi/configure_massa_wallet.go
```diff
@@ -121,7 +121,9 @@ func setupMiddlewares(handler http.Handler) http.Handler {
 // The middleware configuration happens before anything, this middleware also applies to serving the swagger.json document.
 // So this is a good place to plug in a panic handling middleware, logging and metrics.
 func setupGlobalMiddleware(handler http.Handler) http.Handler {
-	handleCORS := cors.Default().Handler
+	handleCORS := cors.New(cors.Options{
+		AllowedMethods: []string{http.MethodGet, http.MethodPost, http.MethodHead, http.MethodDelete},
```
ASAPSegfault, line 125: Are these the only methods we're using in the whole plugin ? 
(No PUT for example ?)

ASAPSegfault (approved):

api/server/restapi/configure_massa_wallet.go
Thykof: for now, no 
image

BatiGencho (approved):

Thykof:
> @Thykof as CORS will need to be added to all plugins, I'm interested to understand why we miss that part yesterday when adding CORS to the wallet plugin? 
> Could you specify the rules here? (to all endpoints? to all functions? ...) 

This PR allow CORS for DELETE method. I thought it was handle by default. 

The general rule is: we need to allow CORS for all the endpoints: routes and methods that our API define

SlnPons:
Please make sure to share this rule internally.
And I'll make sure this point is shared with external devs building plugin via the documentation ;)

Thykof:
> Please make sure to share this rule internally. And I'll make sure this point is shared with external devs building plugin via the documentation ;)

To keep it simple, I think internal developer should refer to the same documentation, the one describe in this issue https://github.com/massalabs/thyra/issues/612

PR #97, opened 2023-04-18
Wails: sign operation

Thykof:


Thykof (commented):
@peterjah you can merge

PR #99, opened 2023-04-19
Improve wails front end

Thykof:


Thykof:
we can close it because it was experimental and now we will how react router

PR #120, opened 2023-04-27
add yml import

peterjah:


internal/handler/wallet/import.go
```diff
@@ -2,15 +2,50 @@ package wallet
...
+	return operations.NewRestWalletImportCreated().WithPayload(
+		&models.Wallet{
+			Nickname: models.Nickname(wallet.Nickname),
+			Address:  wallet.Address,
+			KeyPair: models.WalletKeyPair{
+				PrivateKey: "",
+				PublicKey:  wallet.GetPupKey(),
+				Salt:       "",
+				Nonce:      "",
+			},
```
Thykof, line 49: I think we should reuse code of RestGetWallet

pkg/app/app.go
```diff
@@ -79,3 +84,32 @@ func (a *WalletApp) Show() {
...
+
+	return selectFileResult{FilePath: filePath, Nickname: wallet.Nickname}
+}
+
+func (a *WalletApp) ImportWalletFile(filePath string) {
+	a.WalletFileChan <- filePath
+}
+
+func (a *WalletApp) ImportPrivateKey(pkey string) {
+	fmt.Println("Received password input!")
```
Thykof, line 113: remove debug fmt.Println

```suggestion
```

wails-frontend/wailsjs/go/models.ts
```diff
@@ -0,0 +1,21 @@
+export namespace walletapp {
+	
+	export class selectFileResult {
```
Thykof, line 3: class name shouldn't be capitalized: `SelectFileResult`?

Thykof (changes requested):

wails-frontend/wailsjs/go/models.ts
peterjah: nope, its generated code!

pkg/prompt/import.go
Thykof, line 1: do we need to keep the fmt.Println logs?

wails-frontend/src/pages/importPrivateKey.tsx
Thykof, line 1: there several `if (privateKey)` I don't understand why.
maybe we need to create 2 distinct components?

wails-frontend/src/pages/success.tsx
```diff
@@ -1,14 +1,30 @@
...
+        return 'The account has been deleted';
+      case promptAction.newPasswordReq:
+        return 'The password has been created';
+      case promptAction.importReq:
+        return 'The account has been imported';
+      case promptAction.signReq:
+        return 'The transaction has been signed';
+      default:
+        return 'Apply';
+    }
```
Thykof, line 21: maybe we should start setting up i18n for internationalization and it will simplify all the switch statements like this one?

Thykof (changes requested):
please run `gofumpt -w .` to format go files

pkg/wallet/wallet.go
```diff
@@ -353,9 +353,14 @@ func (w *Wallet) FilePath() (string, error) {
 }
 
 func Import(nickname string, privateKeyB58V string, password string) (*Wallet, error) {
+	// TODO: check if the private key is valid
```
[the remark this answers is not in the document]
peterjah: TODO: create the issue! 😄

pkg/prompt/import.go
peterjah: It is pretty usefull to me but not mandatory... i propose to keep it and remove it before the release. or set a logger with log levels...

wails-frontend/src/pages/success.tsx
peterjah: i don't understand how setup internationalization will simplify this :/

wails-frontend/src/pages/importFile.tsx
```diff
@@ -0,0 +1,67 @@
...
+const ImportFile = () => {
+  const nav = useNavigate();
+
+  const [account, setAccount] = useState<
+    undefined | walletapp.selectFileResult
+  >(undefined);
+  const [errorMsg, setErrorMsg] = useState('');
+
+  const { state } = useLocation();
+  const req: promptRequest = state.req;
```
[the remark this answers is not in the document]
peterjah: state is a customizable object that allow to send data when navigate from a page to another using react router

pkg/wallet/wallet.go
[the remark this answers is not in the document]
peterjah: in the end i removed the todo. because if the key is not valid the decoding will fail just after...

wails-frontend/src/pages/importPrivateKey.tsx
peterjah: it could have been done differently... 2 component or even 2 different pages. if you see an issue with this or if we will benefit to to another way we can change !

wails-frontend/src/pages/success.tsx
Thykof: because instead of the switch you use the translation key. i18n will do the switch for you

t(`prompt.action.${req.Action}`)

wails-frontend/src/pages/importPrivateKey.tsx
Thykof: @mazmassa WDYT?

wails-frontend/src/pages/success.tsx
mazmassa: > i don't understand how setup internationalization will simplify this :/

Neither I! Let's sync and you give more context. In any case is already merged. :)

PR #142, opened 2023-05-10
create-dev-route

pivilartisant:
Created a quick url to enable devs to have a route dedicated for development screens


mazmassa (approved):

peterjah:
@pivilartisant why create a "dev" route instead of the real name ?? like "createaccount" or "home" or something else ?

Thykof:
the route has been added in #143

PR #143, opened 2023-05-10
implement select screen

pivilartisant:
Screenshot from 2023-05-10 11-57-00

This is the select account screen in the wallet onboarding interface


web-frontend/src/pages/Welcome.tsx
```diff
@@ -1,4 +1,5 @@
 import LandingPage from '../layouts/LandingPage/LandingPage';
+import { Button } from '@massalabs/react-ui-kit/src/components/Button/Button';
```
Thykof, line 2: imported but not used
```suggestion
```

web-frontend/src/pages/SelectAccount/SelectAccount.tsx
```diff
@@ -0,0 +1,65 @@
...
+      console.log('clicked');
+    },
+    preIcon: <FiPlus className="w-6 h-6" />,
+  };
+
+  const margin24 = 'mb-6';
+  const margin16 = 'mb-4';
+  const defaultFlex = 'flex flex-col justify-center items-center align-center';
+  const buttonFontColor = 'text-neutral';
+  const selectAccountWidth = 'w-[498px]';
```
Thykof, line 26: I don't see the interest of margin24, margin16, buttonFontColor, selectAccountWidth

Thykof (changes requested):

web-frontend/src/pages/SelectAccount/SelectAccount.tsx
Thykof: WDYT @mazmassa

mazmassa: Agree!

I would say if we have only related we can place it directly to the html tag. Only if we have something bigger or with events like: hover, etc...

mazmassa: BTW, why we are setting that in 498px!? 🤔  Why not 50% or something related!? w-1/2!?

web-frontend/src/pages/SelectAccount/SelectAccount.tsx
```diff
@@ -0,0 +1,65 @@
...
+          <div id="account-select" className="w-full flex flex-col">
+            <div className={margin16}>
+              <AccountSelectorButton {...selectProps} />
+            </div>
+            <div className={margin16}>
+              <AccountSelectorButton {...selectProps} />
+            </div>
+            <div>
+              <Button variant="secondary" {...buttonProps}>
+                {' '}
```
mazmassa, line 56: 🤔

web-frontend/src/pages/Welcome.tsx
pivilartisant: Changed !

web-frontend/src/pages/SelectAccount/SelectAccount.tsx
pivilartisant: Yes ! I wanted to sync with you to see how we change that ! thanks for the suggestion !

web-frontend/src/pages/SelectAccount/SelectAccount.tsx
pivilartisant: it's the linter !

peterjah (approved):

Thykof (approved):

PR #145, opened 2023-05-10
add wallet balance fetch

peterjah:


pkg/network/interface.go
```diff
@@ -0,0 +1,20 @@
...
+
+func NewNodeFetcher() *NodeFetcher {
+	return &NodeFetcher{}
+}
+
+type NodeFetcherInterface interface {
+	GetAccountsInfos(wlt []wallet.Wallet) ([]node.Address, error)
+}
+
+// Verifies at compilation time that WalletPrompter implements WalletPrompterInterface interface.
```
Thykof, line 19:
```suggestion
// Verifies at compilation time that NodeFetcher implements NodeFetcherInterface interface.
```

web/html/wallet.js
```diff
@@ -64,18 +64,19 @@ function createAccount() {
...
+                `/api/accounts/${nickname}`
+            );
+            return resp.data.candidateBalance;
+        } catch (error) {
+            console.error(error)
+            return '-'
+        }
     }
-}
+
```
Thykof, line 79: I don't understand this change.

Thykof (changes requested):

PR #148, opened 2023-05-10
modify tailwind setup

maxime-carabina:
We are modifying tailwind setup.


mazmassa (approved):

Thykof:
npm run dev works without this, not sure it's needed.

PR #158, opened 2023-05-10
add account screen

pivilartisant:
Screenshot from 2023-05-10 16-37-26


web-frontend/package.json
```diff
@@ -17,7 +17,7 @@
     "prepare": "cd .. && husky install web-frontend/.husky"
   },
   "dependencies": {
-    "@massalabs/react-ui-kit": "^0.0.2-dev",
+    "@massalabs/react-ui-kit": "^0.0.4-dev.20230510134734",
```
Thykof, line 20: please run npm i after applying this change.

```suggestion
    "@massalabs/react-ui-kit": "^0.0.4-dev",
```

web-frontend/src/pages/AddAccount/AddAccount.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { FiArrowRight } from 'react-icons/fi';
+import { Button } from '@massalabs/react-ui-kit/src/components/Button/Button';
+
+export default function AddAccount() {
+  const topButtonArgs = {
+    onClick: () => {
+      console.log('clicked');
+    },
+    posIcon: <FiArrowRight />,
+  };
```
Thykof, line 11: @mazmassa why it is better to declare the props in a const here rather than directly where you use the component.

IMO it could be hard to read the code if I have to scroll up to the const declaration while reading the HTML bellow.

peterjah (approved):

Thykof (approved):

web-frontend/src/pages/AddAccount/AddAccount.tsx
mazmassa: Agree!

This may is useful in tests to not set in each one. However in the implementation it would worth put directly in the component, indeed.

mazmassa: However, since this is already merged... maybe a scaling task to update then.

pivilartisant: okay sounds good i'll take note in my future integrations :) @Thykof  @mazmassa

maxime-carabina: Nice recommendation! 👍🏻 
I will create à PR to update this on the ui-kit for my Button component.

PR #161, opened 2023-05-10
fix legacy create account

peterjah:


Thykof (approved):
I don't understand how it fixes the bug, but as it's legacy code I trust you blindly!

PR #168, opened 2023-05-11
Account list

peterjah:


Thykof (approved):

web-frontend/.env.massastation
```diff
@@ -1 +1,2 @@
-VITE_BASE_PATH="/thyra/plugin/massalabs/wallet/web-app/"
\ No newline at end of file
+VITE_BASE_PATH="/thyra/plugin/massalabs/wallet/web-app/"
+VITE_API_URL="/thyra/plugin/massalabs/wallet/api/"
```
mazmassa, line 2: I would suggest to keep all that in an object. 
Most of the time you'll import BASE_API|APP but you can import granularly if you need it too.

```js
export const API = {
  verb: "http",
  namespace: "",
  server: "localhost",
  port: "3001",
};

export const APP = {
  verb: "http",
  namespace: "",
  server: "localhost",
  port: "3000",
};

export const BASE_API = API.port
  ? `${API.verb}://${API.server}:${API.port}`
  : `${API.verb}://${API.server}`;

export const BASE_APP = API.port
  ? `${APP.verb}://${APP.server}:${APP.port}`
  : `${APP.verb}://${APP.server}`;

File: web-frontend/.eslintrc.cjs [hunk 2]
```diff
@@ -1,4 +1,4 @@
 module.exports = {
-  extends: ['@massalabs'],
-  plugins: ['html'],
+  extends: ['@massalabs', 'plugin:@tanstack/eslint-plugin-query/recommended'],
+  plugins: ['html', '@tanstack/query'],
```
mazmassa, line 3: We should keep plugins addition (+package.lock) in one MR.

web-frontend/src/api/account.tsx
```diff
@@ -0,0 +1,6 @@
+import axios from 'axios';
+
+const api = axios.create({
+  baseURL: import.meta.env.VITE_API_URL,
+});
+export const getAllAccounts = () => api.get('accounts').then((res) => res.data);
```
mazmassa, line 6: Can't we use the **useResource** standard we discussed to reach the endpoints!? 🤔

web-frontend/src/main.tsx
```diff
@@ -6,29 +6,31 @@ import {
...
-
 import './index.css';
 import '@massalabs/react-ui-kit/src/global.css';
 import Welcome from './pages/Welcome/Welcome.tsx';
 import SelectAccount from './pages/SelectAccount/SelectAccount.tsx';
 import Error from './pages/Error.tsx';
 import AddAccount from './pages/AddAccount/AddAccount.tsx';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+
+const queryClient = new QueryClient();
```
mazmassa, line 17: This would be in the same specific MR to add the lib. I would be a small MR only concerning the addition.

mazmassa (dismissed):

mazmassa (approved):
It would be better if we add the standards we discussed some week ago. However, I am approving meanwhile to not block you.

We can update in future then. It's up to you to apply now or after! :)

web-frontend/.env.massastation
Thykof: I don't understand your comment, we benefit from vite fonctionnality https://vitejs.dev/guide/env-and-mode.html to make the wallet work for both standalone and as a plugin

PR #169, opened 2023-05-11
adds makefile

Thykof:


Thykof:
I will switch to https://taskfile.dev/

PR #170, opened 2023-05-11
Link welcome page

Thykof:


web-frontend/src/main.tsx
```diff
@@ -13,15 +13,20 @@ import SelectAccount from './pages/SelectAccount/SelectAccount.tsx';
...
+}
 
 const queryClient = new QueryClient();
 
 const router = createBrowserRouter(
   createRoutesFromElements(
-    <Route path={import.meta.env.VITE_BASE_PATH}>
-      <Route path="index" element={<Welcome />} />
+    <Route path={import.meta.env.VITE_BASE_APP}>
+      <Route path="welcome" element={<Welcome />} />
```
peterjah, line 27: keeping "/" or index for the welcome page is nice ? isnt it?

Thykof: it's not possible, goswagger will not fallback into the /web-app/{ressource} route

web-frontend/.env.standalone
```diff
@@ -1,2 +1,4 @@
-VITE_BASE_PATH="/web-app/"
-VITE_API_URL="/api/"
\ No newline at end of file
+VITE_BASE_APP="/web-app"
+VITE_BASE_API="/api"
+# VITE_ENV="dev"
+VITE_ENV="prod"
```
peterjah, line 4: so which one it is ? dev or prod ? :p

web-frontend/src/pages/SelectAccount/SelectAccount.tsx
```diff
@@ -3,59 +3,54 @@ import { FiUser, FiPlus } from 'react-icons/fi';
...
           </div>
           <div className="mb-6">
             <label className="mas-body text-neutral" htmlFor="account-select">
               Select an account
             </label>
           </div>
           <div id="account-select" className="w-full flex flex-col">
-            {accounts.data?.map((account: accountType) => (
-              <div className="mb-4">
+            {data.map((account: AccountObject) => (
```
peterjah, line 38: i think you should keep the "?" after data, it may be undefined in case of http error

peterjah (changes requested):

web-frontend/.env.standalone
Thykof: you choose haha, in dev mod, mirage will catch your API calls. I don't know yet how to use it properly, we will see.

web-frontend/src/pages/AddAccount/AddAccount.tsx
```diff
@@ -23,10 +16,15 @@ export default function AddAccount() {
             <h1 className="mas-banner text-neutral">Add an account</h1>
           </div>
           <div className="mb-4">
-            <Button {...topButtonArgs}>Create an account</Button>
+            <Link to={routeFor('')}>
```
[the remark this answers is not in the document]
Thykof: because the page doesn't exists yet

web-frontend/src/pages/SelectAccount/SelectAccount.tsx
Thykof: normally it's handled with data = [] above

web-frontend/src/main.tsx
Thykof: we can keep as it was if you think index is better than welcome

Thykof:
ok to merge

PR #172, opened 2023-05-12
setup Task

peterjah:


Taskfile.yml
```diff
@@ -0,0 +1,90 @@
...
+    cmds:
+      - cmd: mkdir -p build/wallet-plugin
+      - cmd: go build -tags desktop,production -ldflags "-w -s" -o build/wallet-plugin/wallet-plugin main.go
+    env:
+      CGO_LDFLAGS: -framework UniformTypeIdentifiers
+
+  build-windows:
+    platforms: [windows]
+    cmds:
+      - cmd: go build -tags desktop,production -ldflags "-w -s -H windowsgui" -o build/wallet-plugin/wallet-plugin.exe main.go
```
Thykof, line 44: it will automatically create build/wallet-plugin folder?

Thykof (changes requested):
perfect!

Taskfile.yml
peterjah: good catch! i added mkdir and changed the slashes side.. but its not tested :)

Thykof (approved):

PR #176, opened 2023-05-15
fixed selector import

pivilartisant:
The import for the account selector needed to be fixed after some ui-kit refactoring


peterjah (changes requested):
can you make the changes in react-ui-kit for everything to be exported properly??
I should be possible to import components like this:
`
import { Button, AccountSelector,  MassaToken} from '@massalabs/react-ui-kit';
`

pivilartisant:
> can you make the changes in react-ui-kit for everything to be exported properly?? It should be possible to import components like this: `import { Button, AccountSelector, MassaToken} from '@massalabs/react-ui-kit';`


I wish.... unfortunately we have a problem with imports like that  : 

Screenshot from 2023-05-15 12-19-22

peterjah:
It seems to work fine for Button, so why not for AccountSelector and MassaToken?

Thykof:
it's something related to `import React from 'react';` even if React is not used in the component

pivilartisant:
> It seems to work fine for Button, so why not for AccountSelector and MassaToken?

it works on the go server but not with vite... so I guess best we can do is dev with specific paths and delete them once we are making the pr ? 
@peterjah  @Thykof

peterjah:
> > It seems to work fine for Button, so why not for AccountSelector and MassaToken?
> 
> it works on the go server but not with vite... so I guess best we can do is dev with specific paths and delete them once we are making the pr ? @peterjah @Thykof

sorry i don't understand :/ 
but lets forget about this, its not critical

peterjah (approved):

Thykof:
> > It seems to work fine for Button, so why not for AccountSelector and MassaToken?
> 
> it works on the go server but not with vite... so I guess best we can do is dev with specific paths and delete them once we are making the pr ? @peterjah @Thykof

yes if you want

PR #177, opened 2023-05-15
Preact to react

maxime-carabina:
We are switching the preact wails front-end app to react


wails-frontend/src/events/events.tsx
```diff
@@ -1,6 +1,6 @@
 export type promptResult = {
   Success: boolean;
-  Data: any;
+  Data: [];
```
peterjah, line 3: why do you assume Data is an array ? it is not the case, for now its a string. please let "any" for now. if the linter complais you can disable it for this file!

peterjah (changes requested):

wails-frontend/src/events/events.tsx
maxime-carabina: Ok no problem !

maxime-carabina: Done @peterjah

peterjah (approved):

Thykof:
> what were the motivations?

ui-kit was not importable into preact wail frontend

PR #199, opened 2023-05-16
ci: only build binary for main branch

peterjah:


Thykof (approved):

thomas-senechal:
IMO, we shouldn't remove this `pull_request` trigger for this job. It might help to spot some build errors for other OS's or even just build errors after applying some suggestions from reviews.

For example, I'm working on MacOS. But our code is also supposed to build and run on Linux and Windows, and sometimes, we might do things that won't even build on an other OS. The CI running on pull requests helps to make sure it build correctly and ensure that we won't have to fix it once merged on `main` branch. 

I had this exact issue on https://github.com/massalabs/thyra/pull/657 with Windows not building because I forgot some imports that my IDE didn't notice were missing (branch has been rebased since, so it's not visible).

PR #207, opened 2023-05-16
Add Wails Import User Flow

pivilartisant:
Screenshot from 2023-05-16 12-06-59

### DIFF
--- a/wails-frontend/src/pages/importFile.tsx
+++ b/wails-frontend/src/pages/importFile.tsx
@@ -9,6 +9,7 @@ import { useLocation, useNavigate } from 'react-router-dom';
 import { handleApplyResult, handleCancel } from '../utils/utils';
 import { walletapp } from '../../wailsjs/go/models';
 import { EventsOnce } from '../../wailsjs/runtime/runtime';
+import { Button } from '@massalabs/react-ui-kit';
 
 const ImportFile = () => {
   const nav = useNavigate();
@@ -21,11 +22,15 @@ const ImportFile = () => {
   const { state } = useLocation();
   const req: promptRequest = state.req;
 
+  const accountStyleSuccess = ' mas-body text-s-success  pb-4';
+  const accountStyleNormal = ' mas-body text-neutral  pb-4';
+
   const baselineStr = () =>
     account
-      ? `Importing account ${account.nickname}`
+      ? `Selected ${account.nickname}'s account`
       : 'Select an account file to import';
-  const applyStr = () => (account ? 'Import' : 'Select a file');
+
+  const importStr = () => (account ? 'Import' : 'Select a file');
 
   const handleApply = async () => {
     if (!account) {
@@ -34,7 +39,8 @@ const ImportFile = () => {
         return;
       }
       if (res.err) {
-        setErrorMsg(res.err);
+        // setErrorMsg(res.err);
+        setErrorMsg('The file format must be .yaml');
         return;
       }
       setAccount(res);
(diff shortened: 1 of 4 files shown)


mazmassa (approved):

peterjah (approved):

wails-frontend/src/pages/importFile.tsx
```diff
@@ -48,19 +50,37 @@ const ImportFile = () => {
...
-      <div className="baseline">{baselineStr()}</div>
-      <div className="flex">
-        <button className="btn" onClick={handleCancel}>
-          Cancel
-        </button>
-        <button className="btn" onClick={handleApply}>
-          {applyStr()}
-        </button>
+    <div className="bg-primary flex flex-col justify-center items-center h-screen w-full">
+      <div className="w-1/4  max-w-sm  min-w-fit">
```
mazmassa, line 54: I don't think we need to use **max-w-sm** here.

wails-frontend/src/pages/importFile.tsx
```diff
@@ -48,19 +50,37 @@ const ImportFile = () => {
...
+        <div>
+          <p className="mas-title text-neutral pb-4">{req.Msg}</p>
+        </div>
+        <div>
+          <p
+            className={
+              account
+                ? ' mas-body text-s-success  pb-4'
+                : ' mas-body text-neutral pb-4'
+            }
```
mazmassa, line 64: Can we have that extracted from the code to a const!? 🤔

wails-frontend/src/pages/importMethods.tsx
```diff
@@ -13,18 +14,28 @@ const ImportMethods = () => {
...
-      <div className="baseline">{baselineStr}</div>
-      <div className="flex flex-col">
-        <button className="btn" onClick={handleYml}>
-          I have a .yml file
-        </button>
-        <button className="btn" onClick={handlePkey}>
-          I have a private key
-        </button>
+    <div className="bg-primary flex flex-col justify-center items-center h-screen w-full">
+      <div className="w-1/4  max-w-xs  min-w-fit">
```
mazmassa, line 18: Same here about **max-w-sx**. Please, double check if we need it!

wails-frontend/src/pages/importFile.tsx
```diff
@@ -23,7 +24,7 @@ const ImportFile = () => {
 
   const baselineStr = () =>
     account
-      ? `Importing account ${account.nickname}`
+      ? `Selected ${account.nickname}'s account`
       : 'Select an account file to import';
   const applyStr = () => (account ? 'Import' : 'Select a file');
```
mazmassa, line 29: Why these need to be a function!? 🤔 

Can't we just assign as variable and the in the code access the variable (baselineStr or applyStr)!?

BTW these both names are a mystery... can't we be more explicit what they are!? 🤔

pivilartisant: I have no say, I would recommend seeing that with @peterjah  who set it up !

wails-frontend/src/pages/importFile.tsx
pivilartisant: uhmmm I think we do because the button will be too stretched if we don't have a max-w :) I can attach screen in this conv if necessary

wails-frontend/src/pages/importFile.tsx
pivilartisant: probably !

wails-frontend/src/pages/importMethods.tsx
pivilartisant: see first comment !

wails-frontend/src/pages/importFile.tsx
mazmassa: Please, do it... closing here then!

wails-frontend/src/pages/importFile.tsx
Thykof: give it a try without function, const instead and let'see if it work too and if it's better

pivilartisant: also addressed the naming ! applyStr is now importStr

Thykof: ok for me

wails-frontend/src/pages/importFile.tsx
```diff
@@ -21,11 +22,15 @@ const ImportFile = () => {
   const { state } = useLocation();
   const req: promptRequest = state.req;
 
+  const accountStyleSuccess = ` mas-body text-s-success  pb-4`;
+  const accountStyleNormal = ` mas-body text-neutral  pb-4`;
```
Thykof, line 26:
```suggestion
  const accountStyleSuccess = 'mas-body text-s-success  pb-4';
  const accountStyleNormal = 'mas-body text-neutral  pb-4';
```

Thykof (changes requested):

wails-frontend/src/pages/importFile.tsx
```diff
@@ -48,19 +54,35 @@ const ImportFile = () => {
...
-        </button>
+    <div className="bg-primary flex flex-col justify-center items-center h-screen w-full">
+      <div className="w-1/4  max-w-sm  min-w-fit">
+        <div>
+          <p className="mas-title text-neutral pb-4">{req.Msg}</p>
+        </div>
+        <div>
+          <p
+            className={
+              account ? `${accountStyleSuccess}` : `${accountStyleNormal}`
```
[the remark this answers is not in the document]
pivilartisant: changed

wails-frontend/src/pages/importFile.tsx
pivilartisant: changed

Thykof (approved):

PR #214, opened 2023-05-16
211 bakcend: validate nickname alphanum

Thykof:


peterjah (approved):

SlnPons:
This was not needed.... It's ok to keep it as it has been implemented but for product question, please reach out next time ;)

Thykof:
This PR prevent user to enter special char in the nickname, to implement rule: 

image

SlnPons:
@Thykof can you explain what you have implemented please? 
In the rule I asked to check if the special character error is not a bug so it's not 100% clear to me what you've done:
1. Prevent special characters and so didn't solve the bug
2. Solve the bug and so all characters are accepted for the nickname

Thykof:
I didn't look for a potential bug, I force the nickname to contain only letters, min/maj, number, `-` and `_`, no accent, no special char.

I spoke with @peterjah , there is no bug, but if we want to allow any char, we will have to implement something to make it work.

SlnPons:
@gregLibert we discussed together that there was a bug on thyra wallet and that was the reason special characters were not 

@Thykof I'm really against Devs implementing product rules. What is the rational behind no special char or accent? Does it prevent URL encoding? I don't understand the link between the rule implemented and the URL /API calls

SlnPons:
@Thykof thanks for your explanations. As discussed, the rule you implemented will prevent people from using any special chars which include other alphabetical letters etc... 
So it's good to me. Looking forward to testing it.

PR #219, opened 2023-05-17
Add Failure Screen Wails

pivilartisant:
Screenshot from 2023-05-17 10-47-27

### DIFF
--- a/wails-frontend/src/pages/failure.tsx
+++ b/wails-frontend/src/pages/failure.tsx
@@ -1,0 +1,38 @@
+import { useLocation } from 'react-router';
+import { promptAction, promptRequest } from '../events/events';
+import { FiX } from 'react-icons/fi';
+
+const Failure = () => {
+  const { state } = useLocation();
+  const { req } = state;
+
+  const successMsg = (req: promptRequest) => {
+    switch (req.Action) {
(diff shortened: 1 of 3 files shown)


wails-frontend/src/pages/failure.tsx
```diff
@@ -0,0 +1,38 @@
...
+      case promptAction.deleteReq:
+        return 'The account could not been deleted';
+      case promptAction.newPasswordReq:
+        return 'The password could not be created';
+      case promptAction.importReq:
+        return 'The account could not be imported';
+      case promptAction.signReq:
+        return 'The transaction has been signed';
+      default:
+        return 'Apply';
```
mazmassa, line 20: Is that correct have a default with Apply!? 🤔

mazmassa (approved):
Left a comment but LGTM!

wails-frontend/src/pages/failure.tsx
pivilartisant: no not correct will make changes

Thykof (approved):

wails-frontend/src/pages/failure.tsx
pivilartisant: chenged !

PR #223, opened 2023-05-17
211 error message

Thykof:


pkg/wallet/wallet.go
```diff
@@ -41,6 +42,11 @@ func ErrorAccountNotFound(nickname string) error {
 	return fmt.Errorf("account '%s' not found", nickname)
 }
 
+type WailsError struct {
```
peterjah, line 45: why its called wailsError ? i think the error is not directly related to wails and maybe used also by the web frontend

peterjah (approved):

pkg/wallet/wallet.go
Thykof: this struct describe an error happening in the back-end with a hint about how the front-end must display the error message. For now it's used for wails front-end, that's why I called it this way. 

I'll see if WalletError would better fit

PR #228, opened 2023-05-19
wail app: fix timeout chan

Thykof:
this message in the CtrlChan was never read, so when opening the window again, the chan was already filed, prevent use to do anything

### DIFF
--- a/wails-frontend/vite.config.ts
+++ b/wails-frontend/vite.config.ts
@@ -4,4 +4,7 @@ import react from '@vitejs/plugin-react';
 // https://vitejs.dev/config/
 export default defineConfig({
   plugins: [react()],
+  build: {
+    emptyOutDir: true,
+  },
 });


pkg/app/app.go
```diff
@@ -56,8 +56,7 @@ func (a *WalletApp) BeforeClose(ctx context.Context) bool {
 		return false
 	}
 
-	// Send a cancel message to the prompt and do NOT shutdown
-	a.CtrlChan <- Cancel
```
Thykof, line 60: this message in the CtrlChan was never read, so when opening the window again, the chan was already filed, prevent use to do anything

peterjah: I don't think its that simple.
If a wail window is open, there is chances that a channel is waiting for a password or a cancel message.
I think you are addressing the case where wails is open but nothing is waiting because the prompt has already timeout.
So i think we have to consider the 2 cases, because here you are breaking the first one

peterjah (changes requested):

peterjah (approved):

Thykof (commented):
cool!

PR #231, opened 2023-05-19
211 nickname unique constraint

Thykof:


pkg/wallet/wallet.go
```diff
@@ -439,7 +444,24 @@ func createAccountFromKeys(nickname string, privateKey []byte, publicKey []byte,
...
+	// Load all accounts
+	wallets, err := LoadAll()
+	if err != nil {
+		return false, fmt.Errorf("loading wallets: %w", err)
+	}
+
+	// Check if nickname is unique
+	for _, wallet := range wallets {
+		if wallet.Nickname == nickname {
+			return false, nil
```
peterjah, line 457: maybe you should just return the error, and nil when the nickname is unique.
I would simplify your code because thats makes only 1 check to do then

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/rs/cors v1.8.3
 	github.com/shopspring/decimal v1.3.1
 	github.com/stretchr/testify v1.8.1
-	github.com/wailsapp/wails/v2 v2.4.1
+	github.com/wailsapp/wails/v2 v2.5.0
```
peterjah, line 14: thank you for updating wails!

pkg/wallet/wallet.go
Thykof: I pushed a commit to do this.

internal/handler/wallet/create_test.go
```diff
@@ -58,12 +59,16 @@ func Test_walletCreate_Handle(t *testing.T) {
 
 			assertWallet(t, nickname)
 
-			err = cleanupTestData([]string{nickname})
-			assert.NoError(t, err)
+			defer clean(t, []string{nickname})
```
peterjah, line 62: are you sure about this??
defer will be executed when function is returnning.
Previous version was executing cleanupTestData after each test case loop.

Thykof: for test case 'nickname already exists' I need to have a yaml file so that this test pass. But yeah it's not the best way to do it probably

peterjah (approved):

Taskfile.yml
```diff
@@ -11,13 +11,36 @@ tasks:
...
+      # kill the zombie goroutines
+      - cmd: pgrep -f wallet.test | xargs kill -9 &>/dev/null
+        platforms: [linux, darwin]
+      - cmd: rm -f internal/handler/wallet/wallet_*.yml &>/dev/null
+        platforms: [linux, darwin]
+
+  test: 
+    cmds:
+      - task: clean
+      - cmd: go test -race -v -timeout 500s ./...
```
Thykof, line 25:
```suggestion
      - cmd: go test -v -timeout 60s ./...
```

PR #242, opened 2023-05-25
add backup implem

peterjah:


internal/handler/wallet/backup.go
```diff
@@ -25,23 +25,27 @@ func (w *walletBackupAccount) Handle(params operations.BackupAccountParams) midd
...
+
+	if promptOutput != nil {
+		// for private key backup, send the private key to the wails frontend
+		w.prompterApp.EmitEvent(walletapp.PromptDataEvent,
+			walletapp.EventData{Data: wlt.GetPrivKey()})
 	}
 
-	prompterApp.EmitEvent(walletapp.PromptResultEvent,
+	w.prompterApp.EmitEvent(walletapp.PromptResultEvent,
 		walletapp.EventData{Success: true, CodeMessage: utils.MsgBackupSuccess})
```
Thykof, line 49: here we have potentially 2 events emitted. to make it clearer and avoid us some traps for unit tests, should we emit an event of anther type for the private key

```go
if promptOutput != nil {
		// for private key backup, send the private key to the wails frontend
		w.prompterApp.EmitEvent(walletapp.PromptPrivateKeyEvent,
			walletapp.EventData{Data: wlt.GetPrivKey()})
	}
```

internal/handler/wallet/backup_test.go
```diff
@@ -26,24 +29,110 @@ func Test_walletBackupAccount_Handle(t *testing.T) {
...
+	t.Run("invalid backup method", func(t *testing.T) {
+		testResult := make(chan walletapp.EventData)
+
+		go func() {
+			prompterApp.App().PromptInput <- "unknown method"
+			testResult <- (<-resChan)
+		}()
+
+		resp := backupWallet(t, api, nickname)
+		verifyStatusCode(t, resp, http.StatusUnauthorized)
```
Thykof, line 52: [optional]

here it should be status bad request

optional because it's not the first time we make this mistake

pkg/prompt/prompt.go
```diff
@@ -108,3 +119,11 @@ func InputTypeError(prompterApp WalletPrompterInterface) error {
 
 	return fmt.Errorf(InputTypeErr)
 }
+
+func UserChoiceError(prompterApp WalletPrompterInterface) error {
+	fmt.Println("invalid user choice input")
```
Thykof, line 124:
```suggestion
	fmt.Println(UserChoiceErr)
```

pkg/prompt/prompt.go
```diff
@@ -108,3 +119,11 @@ func InputTypeError(prompterApp WalletPrompterInterface) error {
...
+	prompterApp.EmitEvent(walletapp.PromptResultEvent,
+		walletapp.EventData{Success: false, CodeMessage: UserChoiceErr})
```
Thykof, line 126: I had put all the code messages in the utils package

```suggestion
		walletapp.EventData{Success: false, CodeMessage: utils.ErrUserChoice})
```

internal/handler/wallet/backup.go
```diff
@@ -25,23 +25,27 @@ func (w *walletBackupAccount) Handle(params operations.BackupAccountParams) midd
...
 	if err != nil {
-		return
+		return operations.NewBackupAccountUnauthorized().WithPayload(
+			&models.Error{
+				Code:    errorCanceledAction,
+				Message: "Unable to backup wallet",
+			})
+	}
+
+	if promptOutput != nil {
```
Thykof, line 42: it is not really readable, it checks if the output is not nil and it means that the user requested private key.

internal/handler/wallet/backup_test.go
```diff
@@ -26,24 +29,110 @@ func Test_walletBackupAccount_Handle(t *testing.T) {
 
 	nickname := "walletToBackup"
 	password := "zePassword"
-	_, errGenerate := wallet.Generate(nickname, password)
-	assert.Nil(t, errGenerate)
+	wlt, walletError := wallet.Generate(nickname, password)
+	assert.Nil(t, walletError)
```
Pierregvx, line 33: why is this not err and assert.NoError ?

pkg/prompt/prompt.go
```diff
@@ -60,29 +60,40 @@ func WakeUpPrompt(
...
 				res, keepListening, err = handleNewPasswordPrompt(prompterApp, input)
 			case walletapp.Import:
 				res, keepListening, err = handleImportPrompt(prompterApp, input)
+			case walletapp.Backup:
+				if isPrivateKeyBackup == nil {
+					keepListening, err = handleBackupMethod(prompterApp, input)
+					isPrivateKeyBackup = &keepListening
+				} else {
+					res, keepListening, err = handlePasswordPrompt(prompterApp, input, wallet)
+				}
```
Thykof, line 86: can you add comments, I don't understand why you declare a bool and you only check if it is nil, not true nor false

pkg/app/events.go
```diff
@@ -26,6 +28,7 @@ const (
 	Backup
 	Transfer
 	TradeRolls
+	Uncipher
```
Thykof, line 31: elsewhere, we use the word unprotect

Pierregvx (approved):

Thykof (changes requested):

internal/handler/wallet/backup_test.go
```diff
@@ -26,24 +29,110 @@ func Test_walletBackupAccount_Handle(t *testing.T) {
...
+			// send backup method
+			prompterApp.App().PromptInput <- prompt.PrivateKeyBackup
+			// send password
+			prompterApp.App().PromptInput <- "wrong password"
+
+			result := <-resChan
+
+			checkResultChannel(t, result, false, utils.WrongPassword)
+
+			prompterApp.App().CtrlChan <- walletapp.Cancel
```
Thykof, line 106: warning: we will see in the implementation of wails screens if the wrong password redirect to failure page, I think we will remove this line from here

internal/handler/wallet/backup.go
peterjah: its already the case. there is `PromptDataEvent` and `PromptResultEvent`
But it does not really help for the tests because event id is not used

Thykof (approved):

PR #247, opened 2023-05-26
Link import account

Thykof:


maxime-carabina (approved):

web-frontend/src/pages/Welcome/Welcome.tsx
```diff
@@ -44,7 +44,7 @@ export default function Welcome() {
             <Button
               variant="secondary"
               onClick={() => {
-                handleImport.mutate({} as AccountObject);
+                mutate({} as AccountObject);
```
mazmassa, line 47: Non blocking: 

Are we mutating something empty here!? 🤔 An empty object!? Is that the case!?

Don't we need any kind of file validation in FE at this point? If yes,  I would suggest use a **handleSubmit|Import** function and/or wrap that in a form. This allow us to proceed some validations if needed. 

However if we have not here, ok for me.

web-frontend/src/mirage/server.ts
```diff
@@ -16,6 +16,9 @@ function mockServer(environment = ENV.DEV) {
         nickname() {
           return faker.internet.userName();
         },
+        candidateBalance() {
```
mazmassa, line 19: 🤔 
```suggestion
        userBalance() {
```

peterjah (approved):

web-frontend/src/pages/Welcome/Welcome.tsx
Thykof: it's just a PUT endpoint to wake up wails prompt.

web-frontend/src/mirage/server.ts
Thykof: on the blockchain side, we have final balance and candidate balance, this is why we use the same word in the wallet backend API and also here

web-frontend/src/mirage/server.ts
```diff
@@ -35,6 +38,13 @@ function mockServer(environment = ENV.DEV) {
 
         return schema.find('account', nickname);
       });
+
+      this.put('accounts', (schema) => {
+        return schema.create('account', {
+          nickname: 'imported',
+          candidateBalance: '117',
+        });
+      });
```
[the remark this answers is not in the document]
Thykof: the wallet API is designed like this. this PUT on /accounts is only wake uping the wails prompt to import an account. nothing to update, it will create an new account.

there is no param, this is why I create an new one.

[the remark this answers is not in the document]
mazmassa: Ok! Fair enough then! Mirage is a mirror of the server. If is implemented like that... there's nothing we can mock different then! :)

web-frontend/src/mirage/server.ts
mazmassa: Got it! Thanks!

mazmassa (approved):

PR #249, opened 2023-05-26
web: update step3 and add account

maxime-carabina:
We are updating those 2 screens on web-frontend.

### DIFF
--- a/web-frontend/src/i18n/en_US.json
+++ b/web-frontend/src/i18n/en_US.json
@@ -18,11 +18,14 @@
         "description": "Now, define your password. To maximize security, a pop up will open directly on your device via MassaStation."
       },
       "step3": {
-        "title": "Backup"
+        "title": "Backup",
+        "description": "Security is very important, we recommend backing up your account using a key pair or a .yaml file. Would you like to do it now or later?"
       },
       "buttons": {
         "define_password": "Define a password",
-        "next": "Next"
+        "next": "Next",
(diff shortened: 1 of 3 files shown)


web-frontend/src/i18n/en_US.json
```diff
@@ -18,11 +18,14 @@
...
       "step3": {
-        "title": "Backup"
+        "title": "Backup",
+        "description": "Security is very important, we recommend backing up your account using a key pair or a .yaml file. Would you like to do it now or later?"
       },
       "buttons": {
         "define_password": "Define a password",
-        "next": "Next"
+        "next": "Next",
+        "backup": "Back up",
```
mazmassa, line 27:
```suggestion
        "backup": "Backup",
```

maxime-carabina: One the design, it's "back up".
So what would I do ?

mazmassa: Oh! Check with him then!

peterjah (approved):

web-frontend/src/i18n/en_US.json
Thykof: > Backup is a noun spelled as a single word, while back up is a verb spelled as 2 words

Thykof: here either you skip, or you back up (2 verbs)

PR #250, opened 2023-05-26
Web nickname uniqueness

Thykof:


peterjah (approved):

web-frontend/src/i18n/i18n.ts
```diff
@@ -23,7 +23,11 @@ class I18n {
   public t(key: string): string {
     let copy = this.copy;
     // we are using pick in order to make life easier when the day for plurals and copy with params arrives
-    return dot.pick(key, copy);
+    const result = dot.pick(key, copy);
+    if (result === undefined) {
```
mazmassa, line 27: Are we sure about that!? How a result could be **undefined**!? 🤔

web-frontend/src/validation/nickname.tsx
```diff
@@ -0,0 +1,5 @@
+import { AccountObject } from '../models/AccountModel';
+
+export function isAlreadyExists(nickname: string, accounts: AccountObject[]) {
+  return accounts.some((account) => account.nickname === nickname);
```
mazmassa, line 4: Is the return here a boolean!? 🤔

Thykof: yes

web-frontend/src/i18n/i18n.ts
Thykof: if it's not found. I'd like to show the key if the translation doesn't exists:
- to avoid blank
- to ease debugging

mazmassa: Ah! I see what you mean. All good!
However, if you could do in a simple way:
```tsx

const result = dot.pick(key, copy);
return result ?? key
```

mazmassa (approved):

web-frontend/src/i18n/i18n.ts
Thykof: yes, I'll try this in another PR

PR #251, opened 2023-05-26
Wails backup

peterjah:


wails-frontend/src/pages/backupKeyPairs.tsx
```diff
@@ -53,72 +49,80 @@ function CopyKey() {
...
-      handleApplyResult(navigate, req, setError, isExportAction),
-    );
-
-    return ApplyPassword(password);
+  async function handleResult(result: promptResult) {
+    if (!result.Success) {
+      setError({ error: getErrorMessage(result.CodeMessage) });
+      navigate('/failure', {
+        state: { req },
+      });
```
Thykof, line 77: setError will have no effect on what the user see if we navigate to failure

pkg/prompt/env.go
```diff
@@ -33,5 +39,13 @@ func NewEnvPrompter(app *walletapp.WalletApp) *envPrompter {
 	}
 }
 
+func (w *WalletPrompter) SaveAccountFile(nickname string) (string, error) {
+	return runtime.SaveFileDialog(w.PromptApp.Ctx, runtime.SaveDialogOptions{
+		Title:           "Backup Account File",
+		DefaultFilename: wallet.Filename(nickname),
+		Filters:         []runtime.FileFilter{{DisplayName: "Account File (*.yml)", Pattern: "*.yml"}},
+	})
+}
```
Thykof, line 48: do we need this? this function is never called.

Thykof (changes requested):

web/html/wallet.js
```diff
@@ -62,6 +62,18 @@ function createAccount() {
...
+
+    axios
+        .get(addPrefixUrl(`api/accounts/${nickname}/backup`))
+        .then((resp) => {
+            tableInsert(resp.data);
+            wallets.push(resp.data);
+        })
+        .catch(handleAPIError);
+}
+
```
mazmassa, line 76: Is there any reason to not use our useResource here!? 🤔

Thykof: yes, it's legacy vanilla js, he updated this code to be able to test it quickly

wails-frontend/src/pages/backupKeyPairs.tsx
```diff
@@ -53,72 +49,80 @@ function CopyKey() {
...
     e.preventDefault();
     if (!validate(e)) return;
 
-    save(e);
-    // setShow(true); when backend is ready
-    setShow(false);
+    const formObject = parseForm(e);
+    const { password } = formObject;
+
+    EventsOnce(events.promptResult, handleResult);
```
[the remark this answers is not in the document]
Thykof: ref: https://wails.io/docs/reference/runtime/events/#eventsonce

EventsOnce doesn't return a Promise, so it's sync function, but we need handleSubmit to be async because we await SendPromptInput

web/html/wallet.js
mazmassa: ...but this could be using useResource, no!?

Thykof: no, it's not react, and we will remove this front as soon as the new one cover all its features

mazmassa: Ahh ok! Got it!

mazmassa (approved):
Since this is not the last version, LGTM!

wails-frontend/src/utils/errors.tsx
```diff
@@ -1,19 +1,42 @@
...
+  [ErrorCode.TimeoutError]: 'Timeout error',
+  [ErrorCode.WrongPassword]: 'Wrong password',
+  [ErrorCode.InvalidPromptInput]: 'Invalid user input',
 };
 
-export function getErrorMessage(code: string): string {
-  const errorMessage = errorsEN[code];
+export function getErrorMessage(code: ErrorCode | string): string {
+  if (typeof code === 'string' && !(code in ErrorCode)) {
+    console.log('Unknown error code', code);
```
Thykof, line 27: we are in wails, where can I see this log?

Thykof (approved):

PR #252, opened 2023-05-29
Web nickname format constraint

Thykof:


peterjah (approved):

wails-frontend/src/pages/passwordPrompt.tsx
```diff
@@ -116,6 +123,7 @@ function PasswordPrompt() {
                 defaultValue=""
                 name="passwordConfirm"
                 placeholder="Confirm your password"
```
mazmassa, line 125: We should also use I18n for placeholders. 😇

Thykof: true! there is no i18n in wails yet. I was thinking to add it with @maxime-carabina tomorrow

wails-frontend/src/pages/passwordPrompt.tsx
```diff
@@ -53,20 +53,27 @@ function PasswordPrompt() {
...
+  // error state for the input password confirm field
+  const [errorConfirm, setErrorConfirm] = useState<IErrorObject | null>(null);
 
   function validate(e: SyntheticEvent) {
     const form = parseForm(e);
     const { password, passwordConfirm } = form;
 
+    // reset error states
+    setError(null);
+    setErrorConfirm(null);
```
mazmassa, line 67: Please, double check with @LucasF243 because AFAIK he wants to keep the field in error state (red) until the next submit. At least was agreed in the design system...

Screenshot 2023-05-29 at 15 30 44

wails-frontend/src/pages/passwordPrompt.tsx
mazmassa: Ok! Fair enough! We can create a MR only for that. Good for me!

wails-frontend/src/pages/passwordPrompt.tsx
Thykof: we are in the validate function triggered when user clicks on the submit button. so the error state will not change until the next submit.

Thykof: my question is more: do we show error 'passwords are not the same' bellow the confirm password field even though the password field has error 'must be at least 5 chars'?

mazmassa: Let me see if I got the point: you mean more than one error together or one by one?
If is that you are talking about the way Lucas structured that we should go one by one if we have more than one error 😅 . At least for now!

Thykof: @SlnPons prefers to show all errors, and it easily doable: see my last commit

mazmassa: I am not against, however I would like to understand how this will looks like in terms of UI/UX.

mazmassa: Ohh! Beautiful! 👏 

Now I got your point. For the next time, add that under description of the MR.

Thanks for explaining!

mazmassa (approved):

PR #253, opened 2023-05-29
write and read .yaml, can import .yml

Thykof:
YAML recommend to use .yaml extension.

https://stackoverflow.com/a/22268973/5903242


peterjah (approved):
Just be aware this is breaking change! every previously created/imported wallet will not be detected anymore unless the user change the file extension. maybe it will be better to keep the support of .yml

PR #267, opened 2023-05-30
create backend skeleton

pivilartisant:
We create backend skeleton to start dev


web-frontend/src/main.tsx
```diff
@@ -37,6 +38,7 @@ const router = createBrowserRouter(
       <Route path="account-create-step-one" element={<StepOne />} />
       <Route path="account-create-step-two" element={<StepTwo />} />
       <Route path="account-create-step-three" element={<StepThree />} />
+      <Route path="homepage-transfer-send" element={<SendCoins />} />
```
pivilartisant, line 41: url should be send-coins :)

PR #274, opened 2023-05-31
262 create send operation

Pierregvx:
add simple send-coins page that calls the transfer function

### DIFF
--- a/web-frontend/src/layouts/WalletLayout/WalletLayout.tsx
+++ b/web-frontend/src/layouts/WalletLayout/WalletLayout.tsx
@@ -18,7 +18,8 @@ import { useState } from 'react';
 
 enum MenuItem {
   Home = 'home',
-  SendReceive = 'send-receive',
+  SendCoins = 'send-coins',
+  Receive = 'receive',
   Transactions = 'transactions',
   Contacts = 'contacts',
   Assets = 'assets',
@@ -74,9 +75,12 @@ function WalletLayout(props: WalletProps) {
     {
       label: 'Send/Receive',
       icon: <FiArrowUpRight />,
-      active: isActive(MenuItem.SendReceive),
+      active: isActive(MenuItem.SendCoins),
       footer: false,
-      onClickItem: () => navigate(routeFor(MenuItem.SendReceive)),
+      onClickItem: () =>
+        navigate(routeFor(MenuItem.SendCoins), {
+          state: { nickname, menuItem: MenuItem.SendCoins },
+        }),
     },
     {
       label: 'Contacts',
--- a/web-frontend/src/models/AccountModel.ts
+++ b/web-frontend/src/models/AccountModel.ts
@@ -12,3 +12,8 @@ export type AccountObject = {
   keyPair: keyPairObject;
   nickname: string;
 };
+export type SendTransactionObject = {
+  amount: string;
+  recipientAddress: string;
+  fee: string;
+};
--- a/web-frontend/src/pages/TransferCoins/SendCoins/SendCoins.tsx
+++ b/web-frontend/src/pages/TransferCoins/SendCoins/SendCoins.tsx
@@ -1,21 +1,63 @@
 import { useState } from 'react';
+import usePost from '../../../../custom/api/usePost';
+import { SendTransactionObject } from '../../../../models/AccountModel';
+import { Button } from '@massalabs/react-ui-kit';
+import { useLocation } from 'react-router-dom';
 
 export function SendCoins() {
-  const [amount, setAmount] = useState<number>(64);
-  const [fees, setFees] = useState<number>(42);
-  const [recepient, setRecepient] = useState<number>(42);
+  const { state } = useLocation();
+  const nickname = state.nickname;
+  const { mutate, isSuccess } = usePost<SendTransactionObject>(
+    'accounts',
+    nickname,
+    'transfer',
+  );
+  if (isSuccess) {
+    console.log('Transfer success');
+  }
+
+  const [amount, setAmount] = useState<string>('64');
+  const [fees, setFees] = useState<string>('42');
+  const [recipient, setRecipient] = useState<string>('recipient');
+
+  const handleTransfer = () => {
+    const transferData: SendTransactionObject = {
+      amount: amount,
+      recipientAddress: recipient,
+      fee: fees,
+    };
+
+    mutate(transferData as SendTransactionObject);
+  };
+
+  if (isSuccess) {
+    console.log('Transfer success');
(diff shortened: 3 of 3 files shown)


web-frontend/src/layouts/WalletLayout/WalletLayout.tsx
```diff
@@ -43,7 +44,7 @@ function WalletLayout(props: WalletProps) {
   }
 
   const [selectedItem, setSelectedItem] = useState<MenuItem>(
-    state.menuItem ?? MenuItem.SendReceive,
+    state.menuItem ?? MenuItem.SendCoins,
```
pivilartisant, line 47: ` state.menuItem ?? MenuItem.Home,`

pivilartisant (approved):

web-frontend/src/pages/Home/TransferCoins/SendCoins/SendCoins.tsx
```diff
@@ -1,6 +1,64 @@
...
+import { Button } from '@massalabs/react-ui-kit';
+import { useLocation } from 'react-router-dom';
+import WalletLayout from '../../../../layouts/WalletLayout/WalletLayout';
+export default function SendCoins() {
+  const { state } = useLocation();
+  const nickname = state.nickname;
+  const { mutate, isSuccess } = usePost<SendTransactionObject>(
+    'accounts',
+    nickname,
+    'transfer',
```
mazmassa, line 13: I am not following this! Why are we passing 3 params to usePost here!? 🤔

web-frontend/src/pages/Home/TransferCoins/SendCoins/SendCoins.tsx
```diff
@@ -1,6 +1,64 @@
...
+  );
+  if (isSuccess) {
+    console.log('Transfer success');
+  }
+
+  const [amount, setAmount] = useState<string>('64');
+  const [fees, setFees] = useState<string>('42');
```
mazmassa, line 20: 64? 42? 🤔  Curiosity: what is that exactly!?

web-frontend/src/pages/Home/TransferCoins/SendCoins/SendCoins.tsx
Pierregvx: this is a post query with the url : {{baseURL}}/api/accounts/{{nickname}}/transfer

web-frontend/src/pages/Home/TransferCoins/SendCoins/SendCoins.tsx
Pierregvx: this is just a dummy front that calls the transfer from the back, the real front end will be implemented by @pivilartisant

pivilartisant: placeholder numbers ! I did the form and @Pierregvx  did the backend so we could work with out being stuck

web-frontend/src/pages/Home/TransferCoins/SendCoins/SendCoins.tsx
pivilartisant: because the function usePost take three params : 
`function usePost<T>(
  resource: string,
  nickname: string,
  path?: string,
): UseMutationResult<T, unknown, T, unknown> {
  // All of our routes using POST method as nickname in the url
  var url = `${import.meta.env.VITE_BASE_API}/${resource}/${nickname}`;

  if (path) {
    url = url.concat(`/${path}`);
  }

  return useMutation<T, unknown, T, unknown>({
    mutationKey: [resource],
    mutationFn: async (payload) => {
      const { data } = await axios.post<T>(url, payload);
      return data;
    },
  });
}

export default usePost;
`

pivilartisant: I would recommend pinging @peterjah and/or exploring usePost.ts in thyra-plugin-wallet/web-frontend/src/custom/api/usePost.ts

mazmassa: > because the function usePost take three params :

That's wrong then! It should take only one: **resource**. If the resource is composed by `something/else/more/etc` usePost does not care. All it needs it the resource complete: "something/else/more".

mazmassa: > I would recommend pinging @peterjah and/or exploring usePost.ts in thyra-plugin-wallet/web-frontend/src/custom/api/usePost.ts

No! Here is only FE stuff! usePost is our side. Even in the BE they have these url's

Thykof: yes, here is the curl request:

```
curl --location '127.0.0.1:8080/api/accounts/buildnet/transfer' \
--header 'Content-Type: application/json' \
--header 'Accept: application/json' \
--data '{
    "fee": "10",
    "amount": "10000000",
    "recipientAddress": "AU14nFDPZR5tCkw9Pap6P8Dux2RqAuV6HCee41YVz9ryAgxjH2dX"
}'
```

we needed a optional third argument `path`  for the action /transfer, /sign

web-frontend/src/pages/Home/TransferCoins/SendCoins/SendCoins.tsx
mazmassa: What means placeholder here!? This would be a initial content for the field!? I don't understand...

BTW, you are not aware of the <form> and useState. Let's sync in a call because it's recurrent. Let's clarify that in a call!

web-frontend/src/pages/Home/TransferCoins/SendCoins/SendCoins.tsx
mazmassa: What **nickname** does there!? What if I want to use **usePost** for another regular request!?

Thykof: I don't know, argument nickname was here before adding argument path https://github.com/massalabs/thyra-plugin-wallet/pull/251/files

mazmassa: Ping me if you need help! 

But usePost takes only resource as param. If we need to concatenate something to have the whole resource abc/cde/efg should be before... in the page/component it self.

Thykof:
we merged https://github.com/massalabs/thyra-plugin-wallet/pull/313 instead.

PR #278, opened 2023-05-31
Switch to filename for source of trush of nickname

Thykof:


Thykof:
closing because we modify the standard https://github.com/massalabs/massa-standards/pull/50

PR #280, opened 2023-05-31
Remove b64 logic from the sign function

ASAPSegfault:
Move b64 upwards in code that handles Sign request and replace string parameter of Wallet Sign function by byte array

### DIFF
--- a/pkg/wallet/wallet.go
+++ b/pkg/wallet/wallet.go
@@ -6,8 +6,8 @@ import (
 	"crypto/ed25519"
 	"crypto/rand"
 	"crypto/sha256"
-	"encoding/base64"
 	"fmt"
+	"log"
 	"os"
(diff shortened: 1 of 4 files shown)


pkg/wallet/wallet.go
```diff
@@ -269,6 +268,8 @@ func Load(nickname string) (*Wallet, error) {
 		return nil, fmt.Errorf("getting file path for '%s': %w", nickname, err)
 	}
 
+	log.Print(filePath)
+
```
gregLibert, line 272: to remove ?

gregLibert (changes requested):

Thykof (approved):
I approve once Greg's comment will be solve

gregLibert (approved):

PR #282, opened 2023-05-31
commit swagger generated files

peterjah:


Thykof (approved):
here there are again ^^ let's go

PR #293, opened 2023-06-01
add const stepper in wails

mazmassa:
We are extracting steps in one source of truth.

### DIFF
--- a/wails-frontend/src/i18n/fr_FR.json
+++ b/wails-frontend/src/i18n/fr_FR.json
@@ -7,5 +7,12 @@
     "backup": "Le compte a été sauvegardé",
     "transfer": "La transaction a été envoyée",
     "success": "Succès"
+  },
+  "import": {
+    "steps": {
+      "private": "Clé privée",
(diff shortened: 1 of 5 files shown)


wails-frontend/src/i18n/fr_FR.json
```diff
@@ -7,5 +7,12 @@
     "backup": "Le compte a été sauvegardé",
     "transfer": "La transaction a été envoyée",
     "success": "Succès"
+  },
+  "import": {
+    "steps": {
+      "private": "Private Key",
+      "account": "Account name",
+      "password": "Password"
```
Thykof, line 15: forget French, not priority  for now

```
"private": "Clé privée",
"account": "Nom du compte",
"password": "Mot de passe"
```

Thykof (approved):

wails-frontend/src/i18n/fr_FR.json
mazmassa: Just added! Thanks for that!

PR #301, opened 2023-06-02
300 add update nickname api endpoint

Thykof:


maxime-carabina (approved):

api/walletApi-V0.yml
```diff
@@ -180,19 +180,53 @@ paths:
...
           description: Not found.
           schema:
             $ref: "#/definitions/Error"
-        "500":
+        500:
+          description: Internal Server Error - The server has encountered a situation it does not know how to handle.
+          schema:
+            $ref: "#/definitions/Error"
+    put:
+      operationId: ModifyAccountNickname
```
peterjah, line 200: I think this endpoint should be more generic. do not restrict to nickname. We will probably need to edit some others Account stuffs in the futur
The discription could be something like update account ...

peterjah (changes requested):

peterjah (approved):

PR #308, opened 2023-06-02
300 implement update nickname

Thykof:
during code review, you can ignore the postman collection

### DIFF
--- a/internal/handler/wallet/update.go
+++ b/internal/handler/wallet/update.go
@@ -1,0 +1,89 @@
+package wallet
+
+import (
+	"fmt"
+
+	"github.com/go-openapi/runtime/middleware"
+
+	"github.com/massalabs/thyra-plugin-wallet/api/server/models"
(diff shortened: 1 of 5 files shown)


internal/handler/wallet/update.go
```diff
@@ -0,0 +1,89 @@
...
+	err := wlt.Persist()
+	if err != nil {
+		return nil, &wallet.WalletError{
+			Err:     fmt.Errorf("persisting the modified account: %w", err),
+			CodeErr: utils.ErrAccountFile,
+		}
+	}
+
+	// delete old file
+	wlt.Nickname = oldNickname
```
peterjah, line 77: maybe you can change the DeleteFile func to take the nickname in parameter to avoir doing this weird trick...

peterjah (approved):
Imho the new endpoint should be more generic

internal/handler/wallet/update.go
Thykof: yes good suggestion!

PR #310, opened 2023-06-03
309 change the delete api endpoint so that it return a http response success or failure

Thykof:


internal/handler/wallet/delete_test.go
```diff
@@ -38,38 +40,58 @@ func Test_walletDelete_Handle(t *testing.T) {
 	})
 
 	t.Run("invalid password", func(t *testing.T) {
-		resp := deleteWallet(t, api, nickname)
-
-		verifyStatusCode(t, resp, http.StatusNoContent)
+		// Send password to prompter app and wait for result
+		go func(res chan walletapp.EventData) {
```
peterjah, line 44: the function arg  `res`is not used

internal/handler/wallet/delete_test.go
```diff
@@ -38,38 +40,58 @@ func Test_walletDelete_Handle(t *testing.T) {
...
 		assert.NoError(t, err)
 	})
 
 	t.Run("delete success", func(t *testing.T) {
+		// Send password to prompter app and wait for result
+		go func(res chan walletapp.EventData) {
+			prompterApp.App().PromptInput <- password
+			// forward test result to test goroutine
+			res <- (<-resChan)
+		}(testResult)
```
peterjah, line 86: no need to pass testResult in parameter here. Closures are allowed in go!
`testResult <- (<-resChan)`

peterjah (approved):

PR #321, opened 2023-06-05
bugfix: select the created account in the dropdown

Thykof:
This PR fixes a bug: 
- create a new account
- at step 3, click on Skip
- see in the drop-down account selection that it is not your newly created account

needs https://github.com/massalabs/ui-kit/pull/168

### DIFF
--- a/wails-frontend/package-lock.json
+++ b/wails-frontend/package-lock.json
@@ -1870,9 +1870,9 @@
       }
     },
     "node_modules/@massalabs/react-ui-kit": {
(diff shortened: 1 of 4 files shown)


Thykof:
still to do: clean code, maybe the use effect is not necessary, maybe the options passed to the useRessource is not necessary, also the refetch

Thykof:
indeed, all these changes are useless, we only need https://github.com/massalabs/ui-kit/pull/168

mazmassa (approved):

PR #344, opened 2023-06-06
add support to i18n interpolation key

mazmassa:
We are adding support to interpolation for translation`t()` method.

Now we can have keys written as:

```js
{
  "menu": {
    "home": "what's up {name}! Your user name is {user}",
    "dont": "don't..."
  },
   ...
}
```

to be able to use in the code like:

```ts
Intl.t("menu.home", {name: "NATHAN", user: "NT2343"})
```

### DIFF
--- a/wails-frontend/src/i18n/i18n.ts
+++ b/wails-frontend/src/i18n/i18n.ts
@@ -20,11 +20,14 @@ class I18n {
     this.copy = this._getCopy();
   }
 
-  public t(key: string): string {
+  public t(key: string, interpolations?: Record<string, string>): string {
     let copy = this.copy;
     // we are using pick in order to make life easier when the day for plurals and copy with params arrives
     const result = dot.pick(key, copy);
-    return result ?? key;
+
+    return interpolations
+      ? this._interpolateKeys(result, interpolations)
+      : result ?? key;
   }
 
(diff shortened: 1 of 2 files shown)


Thykof (commented):
I think we should go for the official i18n library instead of reinventing the wheel and duplicating code.

mazmassa:
> I think we should go for the official i18n library instead of reinventing the wheel and duplicating code.

We don't need it for now. In the correct moment we will!

maxime-carabina:
> We are adding support to interpolation for translation`t()` method.
> 
> Now we can have keys written as:
> 
> ```js
> {
>   "menu": {
>     "home": "what's up {name}! Your user name is {user}",
>     "dont": "don't..."
>   },
>    ...
> }
> ```
> 
> to be able to use in the code like:
> 
> ```ts
> Intl.t("menu.home", {name: "NATHAN", user: "NT2343"})
> ```

Oooooh! Nice tips! I didn't know you could do that.
Good news ❤️

maxime-carabina (approved):
LGTM ! 🎉

mazmassa:
> Oooooh! Nice tips! I didn't know you could do that. Good news ❤️

We couldn't! But now yes!

PR #326, opened 2023-06-05
Improve wails messages

Thykof:
image

PR #366, opened 2023-06-08
wails: update sign transaction screen

maxime-carabina:
[duplicate of #326]

We are updating sign transaction screen to follow design : 

image

### DIFF
--- a/internal/handler/wallet/transfer.go
+++ b/internal/handler/wallet/transfer.go
@@ -54,7 +54,7 @@ func (t *transferCoin) Handle(params operations.TransferCoinParams) middleware.R
 
 	promptRequest := prompt.PromptRequest{
 		Action: walletapp.Transfer,
-		Msg:    fmt.Sprintf("Transfer %s nonaMassa from %s to %s, with fee %s nonaMassa", string(params.Body.Amount), wlt.Nickname, *params.Body.RecipientAddress, string(params.Body.Fee)),
+		Msg:    fmt.Sprintf("Transfer %s MAS from %s to %s, with fee %s MAS", string(params.Body.Amount), wlt.Nickname, *params.Body.RecipientAddress, string(params.Body.Fee)),
 	}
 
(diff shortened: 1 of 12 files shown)


mazmassa (approved):

internal/handler/wallet/transfer.go
```diff
@@ -54,7 +54,7 @@ func (t *transferCoin) Handle(params operations.TransferCoinParams) middleware.R
 
 	promptRequest := prompt.PromptRequest{
 		Action: walletapp.Transfer,
-		Msg:    fmt.Sprintf("Transfer %s nonaMassa from %s to %s, with fee %s nonaMassa", string(params.Body.Amount), wlt.Nickname, *params.Body.RecipientAddress, string(params.Body.Fee)),
+		Msg:    fmt.Sprintf("Transfer %s MAS from %s to %s, with fee %s MAS", string(params.Body.Amount), wlt.Nickname, *params.Body.RecipientAddress, string(params.Body.Fee)),
```
Thykof, line 57: Here you change the unit label without changing the value shown. Check if params.Body.Amount is really in MAS and not in nanoMAS (I believe it's nanoMas)

[optional]
I had in mind to pass here an object and format this message string in the frontend:
- in the backend we don't handle internationalization
- it's better to have all texts shown in the same place: en_US.json
I'll try to show you what I had in mind

Thykof: here is what I started: https://github.com/massalabs/thyra-plugin-wallet/pull/326/files but get stuck https://github.com/massalabs/massa-web3/issues/263

Thykof: you can cherry-pick commit https://github.com/massalabs/thyra-plugin-wallet/commit/a5483f93ab7b4a7160981d98271d70327897888f and solve potential conflicts

maxime-carabina: Done

Thykof (approved):

PR #381, opened 2023-06-09
web: add account address section on Home

maxime-carabina:
We are adding account address section on `Home` page.

### DIFF
--- a/web-frontend/src/pages/Home/Home.tsx
+++ b/web-frontend/src/pages/Home/Home.tsx
@@ -1,38 +1,81 @@
-import { useParams } from 'react-router-dom';
+import { useEffect } from 'react';
+import { useNavigate, useParams } from 'react-router-dom';
 import { AccountObject } from '../../models/AccountModel';
-import { formatStandard, Unit } from '../../utils/MassaFormating';
+import { formatStandard, Unit, maskAddress } from '../../utils/MassaFormating';
 import { useResource } from '../../custom/api';
+import { routeFor } from '../../utils';
 import Intl from '../../i18n/i18n';
 
 import WalletLayout, {
   MenuItem,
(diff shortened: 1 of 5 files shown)


web-frontend/src/pages/Home/Home.tsx
```diff
@@ -1,40 +1,63 @@
...
 export default function Home() {
+  const navigate = useNavigate();
   const { nickname } = useParams();
   const { data: account } = useResource<AccountObject>(`accounts/${nickname}`);
 
-  const unformattedBalance = account?.candidateBalance ?? '0';
+  if (account === undefined) {
+    navigate(routeFor(`index`));
+    return null;
+  }
```
mazmassa, line 22: We are relying in the wrong hook. This is not the way. I'll show you just after the daily.

web-frontend/src/pages/Home/Home.tsx
```diff
@@ -1,40 +1,63 @@
...
+            >
               {Intl.t('home.buttons.send')}
             </Button>
           </div>
         </div>
+        <div className="bg-secondary rounded-2xl w-full max-w-lg p-10">
+          <p className="mas-body text-f-primary mb-6">
+            {Intl.t('home.title-account-address')}
+          </p>
+          {/* <CopyContent content={address} formattedContent={formattedAddress} /> */}
```
mazmassa, line 58: 🤔

maxime-carabina: @mazmassa Corrected ! 👍🏻

web-frontend/src/pages/Home/Home.tsx
maxime-carabina: Yes! Let we sync to solve this problem.

PR #384, opened 2023-06-09
add decimal-check-in-inputs

Pierregvx:
image

### DIFF
--- a/web-frontend/src/i18n/en_US.json
+++ b/web-frontend/src/i18n/en_US.json
@@ -70,7 +70,7 @@
     },
     "send": {
       "invalid-amount": "{type} {verb} invalid",
-      "invalid-amount-decimals": "Amount in nMas should be an integer",
+      "invalid-amount-decimals": "{type} can't have decimals",
       "amount-to-high": "Insufficient funds on balance",
       "amount-to-low": "{type} must be greater than 0",
       "no-address": "{type} is required",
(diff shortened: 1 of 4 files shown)


web-frontend/src/i18n/en_US.json
```diff
@@ -70,6 +70,7 @@
     },
     "send": {
       "invalid-amount": "{type} {verb} invalid",
+      "invalid-amount-decimals": "This number has too much decimals : an amount in MAS can't have more than 9 decimals, an amount in nMAS can't have any",
```
mazmassa, line 73:
```suggestion
      "invalid-amount-decimals": "Amount in nMas should be an integer",
```

mazmassa (approved):

pivilartisant (approved):

web-frontend/src/pages/TransferCoins/SendCoins/Send.tsx
```diff
@@ -106,9 +106,10 @@ function Send(props: SendProps) {
   }
 
   function handleFees(num: string) {
+    // check if '.' is present
+    if (num.includes('.')) return;
```
Thykof, line 110: why do you return nothing if . is present? if it's input validation, you should show an error?

Pierregvx: we're blocking the input if a '.' is added
the error is still set in case

PR #394, opened 2023-06-12
web: fix update nickname on settings page

maxime-carabina:
We are fixing deletion wallet when user trying to update his nickname.

### DIFF
--- a/web-frontend/src/pages/Settings/SettingsUpdate.tsx
+++ b/web-frontend/src/pages/Settings/SettingsUpdate.tsx
@@ -1,10 +1,12 @@
-import { useState } from 'react';
-import { useNavigate } from 'react-router-dom';
+import { useState, useRef, SyntheticEvent, useEffect } from 'react';
+import { useNavigate, useParams } from 'react-router-dom';
+import { AxiosError } from 'axios';
 import { usePut } from '../../custom/api';
-import { useParams } from 'react-router-dom';
 import { AccountObject } from '../../models/AccountModel';
 import { routeFor } from '../../utils';
 import Intl from '../../i18n/i18n';
+import { isNicknameValid } from '../../validation/nickname';
+import { parseForm, parseErrors } from '../../utils/index';
 
 import WalletLayout, {
   MenuItem,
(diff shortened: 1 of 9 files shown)


web-frontend/src/main.tsx
```diff
@@ -25,7 +25,7 @@ import SendCoins from './pages/TransferCoins/SendCoins.tsx';
 import Home from './pages/Home/Home.tsx';
 import ReceiveCoins from './pages/TransferCoins/ReceiveCoins/ReceiveCoins.tsx';
 import Settings from './pages/Settings/Settings.tsx';
-import SettingsEdit from './pages/Settings/SeetingsEdit.tsx';
+import SettingsEdit from './pages/Settings/SettingsUpdate.tsx';
```
mazmassa, line 28: SettingsEdit -> SettingsEdit ✅ 

now you are changing to:

SettingsEdit -> SettingsUpdate 🚫 

🤯

web-frontend/src/models/status.ts
```diff
@@ -0,0 +1,4 @@
+export const BAD_REQUEST = 400;
+export const NOT_FOUND = 404;
+export const UNPROCESSABLE_ENTITY = 422;
+export const INTERNAL_SERVER_ERROR = 500;
```
mazmassa, line 4: 👏

web-frontend/src/pages/Settings/SettingsUpdate.tsx
```diff
@@ -0,0 +1,121 @@
...
+        setError({ message: Intl.t('errors.account-not-found') });
+        break;
+      case UNPROCESSABLE_ENTITY:
+        setError({ message: Intl.t('errors.account-already-exists') });
+        break;
+      default:
+        setError({ message: Intl.t('errors.unexpected-error.description') });
+        break;
+    }
+  }
```
mazmassa, line 59: Humm, you can't rely all BAD_REQUEST, and/or NOT_FOUND and etc to map to these errors. A BAD_REQUEST and etc can be more things than only invalid format.

web-frontend/src/main.tsx
maxime-carabina: Oh sorry I miss that

web-frontend/src/pages/Settings/SettingsUpdate.tsx
maxime-carabina: > Humm, you can't rely all BAD_REQUEST, and/or NOT_FOUND and etc to map to these errors. A BAD_REQUEST and etc can be more things than only invalid format.

SO what can I do ? 🤔

web-frontend/src/main.tsx
maxime-carabina: Done

web-frontend/src/pages/Settings/SettingsUpdate.tsx
maxime-carabina: Corrected!
But we have to check if backend return good code (500, 422, etc...) in another PR.

PR #416, opened 2023-06-15
add routefor in replaceState

Pierregvx:


Thykof:
why is it necessary?

Pierregvx:
> why is it necessary?

routeFor adds the vite_base_app to the root of the url

Thykof (approved):

PR #433, opened 2023-06-19
Update author in manifest.json

Pierregvx:


Thykof (approved):

lilyanB (approved):

Pierregvx:
> I can't test this as I can't see the Author in the front-end. So for now it's KO to me...

we need to do a new release

PR #438, opened 2023-06-21
Update manifest.json

pivilartisant:
We change the manifest to return correct plugin name


Thykof (approved):

SlnPons:
@pivilartisant can't see new info in Massa Station

Thykof:
maybe you have to delete the wallet plugin, then see it in the available ones

PR #468, opened 2023-06-27
467 push generated files

ASAPSegfault:


Thykof:
Why did you need that? to be able to import station-massa-wallet as a go package? We can push goswagger generated files for sure, but we need to find another trick for frontend generated file: push a .gitkeep file for example

Thykof:
- I pushed go swagger generated files
- I added .gitkeep file where go embed need files

The problem is that when we generate the frontends, it empty the dist directory, hence it remove the .gitkeep file.

PR #481, opened 2023-06-29
wallet: fix account creation error message in step1

maxime-carabina:
Fix account create error message in step1 when the user tried to enter a name that had already been used.

**BEFORE**
image

**AFTER**
image

**For PM**
**Time estimation :** 10min
**Time result :** 10min

### DIFF
--- a/web-frontend/src/pages/CreateAccount/StepOne.tsx
+++ b/web-frontend/src/pages/CreateAccount/StepOne.tsx
@@ -34,7 +34,7 @@ export default function StepOne() {
     }
 
     if (isAlreadyExists(nickname, accounts)) {
-      setError({ nickname: Intl.t('errors.nickname-already-exists') });
+      setError({ nickname: Intl.t('errors.duplicate-nickname') });
       return false;
     }
 


api/server/models/sign_request.go
```diff
@@ -92,6 +92,10 @@ func (m *SignRequest) ContextValidate(ctx context.Context, formats strfmt.Regist
 
 func (m *SignRequest) contextValidateCorrelationID(ctx context.Context, formats strfmt.Registry) error {
 
+	if swag.IsZero(m.CorrelationID) { // not required
```
Pierregvx, line 95: should'nt this be in the gitignore ?

Thykof (commented):
Could you uncommit changes of .go file please and share with use your version of go?

Pierregvx (approved):
LGTM

Thykof (approved):

api/server/models/sign_request.go
maxime-carabina: Fixed

mazmassa (approved):

PR #482, opened 2023-06-29
Add Dynamic Address Render

pivilartisant:
BEFORE : 
Screenshot from 2023-06-29 11-38-14
 = static address 

AFTER : 

Screenshot from 2023-06-30 17-41-29

LEGACY : 
[Screencast from 29-06-2023 11:38:33.webm](https://github.com/massalabs/station-massa-wallet/assets/90157528/19346970-589c-45ff-a158-875f7ed9dfc5)
 = address render on screen bigger than 1920px 

Update : this feature demands the use of useEffect react hooks which are not optimal and may introduce bugs. Moreover, the current solution has a complexity of O(n) which is not optimal if we want to apply this effect elsewhere in our application. 

After meeting with mario, we have a proposition for you : 

Make the address string adaptable to the size of the object field using css with a complexity of O(1). Here is and example of the results :
Screenshot from 2023-06-29 16-09-57
Screenshot from 2023-06-29 16-10-08

This does not respect design or respect standards so it is a proposition, feel free to give us feedback on this. If it is KO, I will search for a compromise between optimal code and product standards.

### DIFF
--- a/web-frontend/src/pages/Home/Home.tsx
+++ b/web-frontend/src/pages/Home/Home.tsx
@@ -1,4 +1,4 @@
-import { useEffect } from 'react';
+import { useEffect, useState } from 'react';
 import { useNavigate, useParams } from 'react-router-dom';
 import { AccountObject } from '../../models/AccountModel';
 import { formatStandard, Unit, maskAddress } from '../../utils/massaFormat';
@@ -23,6 +23,22 @@ export default function Home() {
     isLoading,
   } = useResource<AccountObject>(`accounts/${nickname}`);
 
+  const breakpoint = 1920; // Screen breakpoint value
+  const unformattedBalance = account?.candidateBalance ?? '0';
+  const balance = parseInt(unformattedBalance);
+  const formattedBalance = formatStandard(balance, Unit.NanoMAS);
+  const address = account?.address ?? '';
+  const formattedAddress = maskAddress(address);
+
+  const [screenWidth, setScreenWidth] = useState(window.innerWidth);
+  const [displayedContent, setDisplayedContent] = useState(formattedAddress);
+
+  // Function seems unnecessary,
+  // but it is needed to update the state of screenWidth
+  function handleResize() {
+    setScreenWidth(window.innerWidth);
+  }
+
   useEffect(() => {
     if (error) {
       navigate(routeFor('error'));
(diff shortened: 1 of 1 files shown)


web-frontend/src/pages/Home/Home.tsx
```diff
@@ -31,11 +47,20 @@ export default function Home() {
...
   }, [account, navigate]);
 
-  const unformattedBalance = account?.candidateBalance ?? '0';
-  const balance = parseInt(unformattedBalance);
-  const formattedBalance = formatStandard(balance, Unit.NanoMAS);
-  const address = account?.address ?? '';
-  const formattedAddress = maskAddress(address);
+  // Call handleResize once after component mounts to log formattedAddress on page load
+  useEffect(() => {
+    window.addEventListener('resize', handleResize);
```
mazmassa, line 52: Please, don't!

web-frontend/src/pages/Home/Home.tsx
```diff
@@ -31,11 +47,20 @@ export default function Home() {
...
+    return () => {
+      window.removeEventListener('resize', handleResize);
+    };
+  }, []);
+
+  useEffect(() => {
+    screenWidth <= breakpoint
+      ? setDisplayedContent(formattedAddress)
+      : setDisplayedContent(address);
+  }, [screenWidth, breakpoint, address, formattedAddress]);
```
mazmassa, line 63: TBH, I don't think we need all that to deal with this feature.

mazmassa:
INHO these changes should be addressed in the clipboard on the Ui-Kit.

web-frontend/src/pages/Home/Home.tsx
pivilartisant: it is a P1

web-frontend/src/pages/Home/Home.tsx
pivilartisant: can you explain more why this is not good ?

web-frontend/src/pages/Home/Home.tsx
mazmassa: I am not saying you don't need to address it. The way you are doing is just not necessary. You can deal with in the ui-kit to be valid for all components and using (I guess, only css).

Why introduce 2 useEffects with listeners!?? 🤔

pivilartisant:
> IMHO these changes should be addressed in the clipboard on the Ui-Kit.

My first plan was to make the changes in the UI-KIT, but "formatted address" is formatted in home.tsx so I am dependent on this factor to conditionally render it. 
  
That being said I am open to rework propositions : how complex in terms of time & rework would you evaluate the change to be ?

web-frontend/src/pages/Home/Home.tsx
mazmassa: Let's sync in a call...

pivilartisant:
@SlnPons I have updated the description of this MR. Please take a look at it and give me you feedback !

web-frontend/src/pages/Home/Home.tsx
gregLibert: And for those that were not in the call, what is the solution ? What is the bad practice ?

SlnPons:
@pivilartisant @mazmassa thanks a lot for trying and for the workout. 
The suggested solution `Make the address string adaptable to the size of the object field using css with a complexity of O(1).`   is ok.

But would it be possible to see the address in full when you over the field? (nice to have)

web-frontend/src/pages/Home/Home.tsx
pivilartisant: it's not a scalable solution with a complexity of O(n) if we want to apply it elsewhere + useEffect hook is prone to introducing uncontrolled bugs that can be hard to identify. @SlnPons validated an alternative. All the information is in the pr description

pivilartisant:
> @pivilartisant @mazmassa thanks a lot for trying and for the workout. The suggested solution `Make the address string adaptable to the size of the object field using css with a complexity of O(1).` is ok.
> 
> But would it be possible to see the address in full when you over the field? (nice to have)
I will create a seperate PR :)

Pierregvx (approved):

maxime-carabina (approved):

PR #486, opened 2023-06-29
add condition on transfer between accounts

Pierregvx:


web-frontend/src/pages/TransferCoins/SendCoins/ContactList.tsx
```diff
@@ -7,16 +7,10 @@ import {
   MassaLogo,
 } from '@massalabs/react-ui-kit';
 import { formatStandard } from '../../../utils/massaFormat';
-import { useResource } from '../../../custom/api';
 import { AccountObject } from '../../../models/AccountModel';
 
 function AccountSelect({ ...props }) {
```
Thykof, line 12: can you create an interface for AccountSelect porps?

web-frontend/src/pages/TransferCoins/SendCoins/ContactList.tsx
```diff
@@ -7,16 +7,10 @@ import {
...
-  const { onClose, setRecipient, account: currentAccount } = props;
-
-  const { data: accounts = [] } = useResource<AccountObject[]>('accounts');
-  const filteredAccounts = accounts.filter(
-    (account: AccountObject) => account.nickname !== currentAccount.nickname,
-  );
+  const { onClose, setRecipient, filteredAccounts } = props;
```
Thykof, line 13: why are you renaming the prop account to filteredAccounts? if the filter is apply in the parent component, then this component is not aware of the filter, and so account is a good variable name

Thykof (changes requested):
Can you create an interface for formObject?

web-frontend/src/pages/TransferCoins/SendCoins/ContactList.tsx
Pierregvx: i kept the same naming but it si true that it is better with accounts now :+1:

web-frontend/src/pages/TransferCoins/SendCoins/ContactList.tsx
```diff
@@ -7,16 +7,11 @@ import {
   MassaLogo,
 } from '@massalabs/react-ui-kit';
 import { formatStandard } from '../../../utils/massaFormat';
-import { useResource } from '../../../custom/api';
 import { AccountObject } from '../../../models/AccountModel';
+import { ContactListProps } from './SendForm';
```
Thykof, line 11: it could be better if ContactListProps in declared inside file ContactList.tsx

Thykof (approved):

web-frontend/src/pages/TransferCoins/SendCoins/ContactList.tsx
```diff
@@ -7,16 +7,16 @@ import {
   MassaLogo,
 } from '@massalabs/react-ui-kit';
 import { formatStandard } from '../../../utils/massaFormat';
-import { useResource } from '../../../custom/api';
 import { AccountObject } from '../../../models/AccountModel';
 
-function AccountSelect({ ...props }) {
-  const { onClose, setRecipient, account: currentAccount } = props;
+export interface ContactListProps {
```
maxime-carabina, line 12:
```suggestion
interface ContactListProps {
```

**TIPS:** if you don't use this interface outside this file, remove `export` flag 👍🏻

web-frontend/src/pages/TransferCoins/SendCoins/ContactList.tsx
```diff
@@ -36,7 +36,7 @@ function AccountSelect({ ...props }) {
       </PopupModalHeader>
       <PopupModalContent>
         <div className="overflow-scroll h-80">
-          {filteredAccounts.map((filteredAccount: AccountObject) => (
+          {accounts.map((filteredAccount: AccountObject) => (
```
maxime-carabina, line 39:
```suggestion
          {accounts.map((filteredAccount: AccountObject, index: number) => (
```

`.map` method always have a `key`. So, can you add an `index` value and put this `index` in `Selector` props ? 🤔 
Like this :

```typescript
<Selector
    customClass="pb-4"
    key={index}
    preIcon={<Identicon username={filteredAccount.nickname} />}
    posIcon={<MassaLogo size={24} />}
    content={filteredAccount.nickname}
    variant="secondary"
    amount={formatStandard(
      +filteredAccount.candidateBalance / 10 ** 9,
    )}
    onClick={() => handleSetRecipient(filteredAccount)}
    />
```

web-frontend/src/pages/TransferCoins/SendCoins/SendForm.tsx
```diff
@@ -37,7 +39,10 @@ export function SendForm({ ...props }) {
   );
   const [fees, setFees] = useState(data?.fees ?? '1000');
   const [recipient, setRecipient] = useState(data?.recipient ?? '');
```
maxime-carabina, line 41:
```suggestion
  const [fees, setFees] = useState<string>(data?.fees ?? '1000');
  const [recipient, setRecipient] = useState<string>(data?.recipient ?? '');
```

maxime-carabina (commented):
```typescript
function validate(formObject: any) {
``` 

You have a function without type specified in params.
Can you add his type ? 🤔 

**TIPS :** If it's an object, add corresponding type object.

web-frontend/src/pages/TransferCoins/SendCoins/ContactList.tsx
Pierregvx: good catch :) I was using it but I moved it

web-frontend/src/pages/TransferCoins/SendCoins/SendForm.tsx
Pierregvx: Yes I saw that and I figured I would do a pr for it but I guess I can do it here

maxime-carabina: > Yes I saw that and I figured I would do a pr for it but I guess I can do it here

Ye! You can do it here if you want because it's just 2 line 😉

maxime-carabina (approved):
LGTM 🎉

PR #487, opened 2023-06-29
Prevent Enter key showing password

pivilartisant:
On enter key press password reveal is now prevented. 

current state : 

As a user, when I am signing a transaction using the wails pop-up, I want to click on the "enter" key of my keyboard and that nothing happens.

Objective : 
User flow
As a user, when I am signing a transaction using the wails pop-up, I want to be able to click on the "enter" key of my keyboard and have the action "Sign" selected and pressed.


maxime-carabina:
> problem is solved here : [massalabs/ui-kit#276](https://github.com/massalabs/ui-kit/pull/276)

For more information, please click on the following link: https://stackoverflow.com/a/33212911/20425931

Thykof (approved):

pivilartisant:
closing because it was not the best approach

Issue #444, opened 2023-06-21
Automatically update the version of the plugin in the manifest.json


SlnPons:
Context

The plugin manifest should always be up to date as it is used on Massa Station module store so if the info isn't correct it could prevent end-users to access the plugin.
In this Manifest, one of the info that changes the most is: the version. We should automate the update of the version based on the release number.

User flow

As a dev, each time a new release is published as "latest", I want to see the "version" field of manifest.json of the plug-in updated with the latest release version tagged as "latest"

How to

List the step-by-step to get it do if needed

Technical details

Give the technical insights so anyone in the team can tackle the tasks

QA testing

Does this task require some QA tests ?
If yes, explain how to validate it

Thykof:
I think there is something wrong in the user flow. The version on the manifest that the Massa Station plugin manager is the one in the release. Because the plugin store gives the URL of the release assets. So the version must be up to date before creating the assets, so before creating the release. 

So we must be more precise in the user flow: 
- I publish version 0.0.n
- the CI update to 0.0.n. (this i what #502 do) but the assets of release of 0.0.n isn't aware of this. it's probably 0.0.n-1, so it's wrong.

This user flow is also wrong:
- I set the version in the manifest to 0.0.k
- I publish version 0.0.k
- the CI updates to 0.0.k+1 (this is not what #502 do)
- I publish version 0.0.k+1: all good the version in the assets is right.
- the CI updates to 0.0.k+2
- I publish version 0.1.0: error, the asset is a 0.0.k+2

The CI can't know in advance what version you will release, (it is not always +1).

SlnPons:
@B-Naoufal please make sure to apply this way of thinking and doing when starting a ticket to avoid wasting time. Also when you see such a comment and that it's blocking you, please ping me so we can discuss, decide and unblock asap :)

@Thykof thanks a lot for your analysis! It's pretty clear. Could you check my logic please?
Here we're just trying to optimise the wallet repo to automate one step: updating the manifest.json of the wallet. 
Why? Because PM is in charge of pushing the release button but it's not that straightforward to update the manifest (even tho I could agree that it's not really complicated either...). So I thought let's automate this step. 
As far as I am aware of in the manifest.json we don't store the assets so this part is still possible.
**NB: the trigger here must be the flag "latest" and not "0.0.n+1" as we can do pre-release or other things like this to test our product.**

However, on Massa Station side, what will happen when the manifest is updated?
[...]

PR #502, opened 2023-07-05
update manifest automatically

B-Naoufal:


pivilartisant (approved):

Thykof (changes requested):
I think this will not solve our need, see https://github.com/massalabs/station-massa-wallet/issues/444#issuecomment-1625072218

Thykof:
I am trying to test this. It doesn't work, I create a branch here https://github.com/massalabs/station-massa-wallet/tree/444-nathan but I am stuck: it creates a commit but the build workflow doesn't take it into account.

Thykof:
So I have the idea to:
- a job that update the manifest, commit and push
- adding a step in the build job to update the manifest, with action https://github.com/marketplace/actions/os-specific-run and for sed on windows: with https://github.com/mbuilov/sed-windows/releases

@peterjah WDYT?

peterjah:
> So I have the idea to:
> 
> * a job that update the manifest, commit and push
> * adding a step in the build job to update the manifest, with action https://github.com/marketplace/actions/os-specific-run and for sed on windows: with https://github.com/mbuilov/sed-windows/releases
> 
> @peterjah WDYT?
creating commits from CI is always a pain.. What i will do is just set the version in manifest before uploading release assets and leave a template value...
But most of the time the most efficient process is just to do the bump commit/or release commit to set the version manually...

Thykof:
closing in favor of https://github.com/massalabs/station-massa-wallet/pull/626

PR #506, opened 2023-07-06
Format address

pivilartisant:
Screenshot from 2023-07-06 15-12-48

Screenshot from 2023-07-06 15-12-52

### DIFF
--- a/wails-frontend/src/pages/backupKeyPairs.tsx
+++ b/wails-frontend/src/pages/backupKeyPairs.tsx
@@ -117,6 +117,7 @@ function BackupKeyPairs() {
                 <Clipboard
                   displayedContent={publicKey}
                   rawContent={publicKey as string}
+                  toggleHover={false}
                 />
               </div>
               <div>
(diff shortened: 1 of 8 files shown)


Pierregvx:
can you show a screenshot ?

wails-frontend/src/pages/backupKeyPairs.tsx
```diff
@@ -117,6 +117,7 @@ function BackupKeyPairs() {
                 <Clipboard
                   displayedContent={publicKey}
                   rawContent={publicKey as string}
+                  toggleHover={false}
```
maxime-carabina, line 120: For what you set `toggleHover` to `false` ?

pivilartisant: @SlnPons and I  (for multiple product related reasons) have decided to remove the toggle feature, but keep it in the code base. So we only remove it from components but it stays in the ui-kit.

maxime-carabina: > @SlnPons and I (for multiple product related reasons) have decided to remove the toggle feature, but keep it in the code base. So we only remove it from components but it stays in the ui-kit.

OK, but if you decide to set it to Off, that means that there is no default value in the UI-KIT.

Unless you've made sure that when you don't use the 'toggleHover' value, there's no hover.

Pierregvx (approved):

wails-frontend/src/pages/backupKeyPairs.tsx
maxime-carabina: We will do this update in a new PR

maxime-carabina (approved):

PR #519, opened 2023-07-11
wallet: migrate mirage from js to ts

maxime-carabina:
We are updating `mirage` from `JS` to `TS`.


web-frontend/src/mirage/factories/account.ts
```diff
@@ -0,0 +1,26 @@
...
+    return faker.internet.userName();
+  },
+  candidateBalance() {
+    return faker.number.int().toString();
+  },
+  balance() {
+    return faker.number.int().toString();
+  },
+  address() {
+    return 'AU' + faker.string.alpha({ length: { min: 128, max: 256 } });
```
Pierregvx, line 16: not blocking but if I am not wrong an address is 53 char max

web-frontend/src/mirage/factories/account.ts
```diff
@@ -0,0 +1,26 @@
...
+  },
+  keyPair() {
+    return {
+      privateKey: faker.string.alpha({ length: { min: 128, max: 256 } }),
+      publicKey: faker.string.alpha({ length: { min: 128, max: 256 } }),
+      nonce: faker.lorem.word(),
```
Pierregvx, line 22: isn't nonce a number ?

Pierregvx (approved):

web-frontend/src/mirage/factories/account.ts
Pierregvx: in case we use it in the front I think we should set it as a number.toString()

maxime-carabina: I just copy the same factory as in the past. But it's a good question 🤔

web-frontend/src/mirage/factories/account.ts
maxime-carabina: No! Address can be long but we can check that with @Thykof  🤔

Thykof: for the sake of the faker, min 50 max 53

web-frontend/src/mirage/factories/account.ts
maxime-carabina: > see here
> 
> https://github.com/massalabs/station-massa-wallet/blob/8507322fb6d9ea85d6a1f0fbe15d060163a94cc7/api/walletApi-V0.yml#L95
> 
> it depends on the ciphered query param: sensitive data (private key, nonce, salt) are returned only if ciphered if false. It's not words, it's base58CheckEncoded binary data: https://github.com/massalabs/station-massa-wallet/blob/6ca35c3040b3746ec56fd6fb0e714abdae61f76c/pkg/wallet/wallet.go#L529C44

So.... what can I do with faker ?

Thykof: empty strings ^^ 

It depends what we want to simulate: backup private key? then we need a fake private key in the body response

maxime-carabina: > empty strings ^^
> 
> It depends what we want to simulate: backup private key? then we need a fake private key in the body response

Ok! I will put an empty string for now 😉

Thykof: to reflect the API all sensitive data (private key, nonce, salt) should be empty string or random string

Thykof (approved):

PR #521, opened 2023-07-11
web: add absolute path

maxime-carabina:
We are adding absolute path in web-frontend.

**EX:** 
```typescript
-- BEFORE --
import { Component } from '../../components';

-- AFTER -- 
import { Component } from '@/components';
```


pivilartisant (approved):
Ty for this !

Pierregvx (approved):
LGTM, thanks !

PR #526, opened 2023-07-13
add signature encoding

ASAPSegfault:


pkg/wallet/wallet.go
```diff
@@ -450,7 +451,9 @@ func createAccountFromKeys(nickname string, privateKey []byte, publicKey []byte,
 		return nil, &WalletError{fmt.Errorf("invalid nickname"), utils.ErrInvalidNickname}
 	}
 
-	address := addressFromPublicKey(publicKey)
+	versionedPubKey := append([]byte{PubKeyVersion}, publicKey...)
+
+	address := addressFromPublicKey(versionedPubKey)
```
Thykof, line 456: modify directly `addressFromPublicKey`: this function should call `wallet.VersionedPubKey()`

Thykof (changes requested):

pkg/wallet/wallet.go
ASAPSegfault: No because wallet.VersionedPubKey() read the public key located inside the wallet, but in this function nothing has been persisted yet, so this function would only return the version byte and nothing else (as I understand things)

Thykof: let's merge and will see once merge how we can refactor all that

Thykof (approved):

PR #529, opened 2023-07-13
skip invalid yaml in load all accounts

Thykof:
Imagine in your wallet plugin folder, you have your YAML file, and you have another YAML file that is either poorly formatted or encrypted incorrectly (an old version...). In this case, today, the wallet plugin doesn't work at all because it doesn't display any of your accounts (even the ones that work). You need to manually delete the problematic YAML file. I suggest ignoring all the YAML files that cause problems but still displaying the ones that work so that you can use the plugin.


B-Naoufal (approved):
I agree.

PR #530, opened 2023-07-13
Update to testnet23

Thykof:


pkg/wallet/wallet.go
```diff
@@ -519,18 +544,102 @@ func AddressIsUnique(address string) error {
...
+	if len(pubKey) == ed25519.PublicKeySize {
+		return append([]byte{PubKeyVersion}, pubKey...)
+	}
+
+	return pubKey
+}
+
+// RemovePubKeyVersion removes the version byte of a public key if the public key has a version byte.
+func RemovePubKeyVersion(versionedPubKey []byte) []byte {
+	if len(versionedPubKey) == ed25519.PublicKeySize+1 {
```
peterjah, line 569: is there any expected cases where  this can be false ? if not you should just check the lenght prior to that and return an error if it is not the expected one

Thykof: you are right I reworked this part

Thykof: https://github.com/massalabs/station-massa-wallet/pull/530/commits/29a383abdc7c24b4e69f8b1130664fe267f7fee4

pkg/wallet/wallet.go
```diff
@@ -31,10 +31,15 @@ const (
...
 	Base58Version             = 0x00
+	AddressVersion            = 0x00
+	PubKeyVersion             = 0x00
+	SignatureVersion          = 0x00
+	PrivKeyVersion            = 0x00
 	UserAddressPrefix         = "AU"
 	PublicKeyPrefix           = "P"
 	PrivateKeyPrefix          = "S"
 	MaxNicknameLength         = 32
+	AccountVersion            = 1
```
gregLibert, line 42: Other versions are in hexa, why not this one ?

pkg/wallet/wallet.go
```diff
@@ -521,18 +546,82 @@ func AddressIsUnique(address string) error {
...
+	knownVersions := []byte{PubKeyVersion}
+
+	for _, version := range knownVersions {
+		if versionedPubKey[0] == version {
+			return versionedPubKey, nil
+		}
+	}
+
+	return nil, fmt.Errorf("invalid version byte")
 }
```
gregLibert, line 579: You should create a VersionKey object with different methods to improve code clarity and avoid duplicating the code for public keys and private keys.

Something like:
```golang
type VersionedKey byte[]

func (v VersionKey) AddVersion(version int) []byte {
  return append([]byte{version}, v...)
}

func (v VersionKey) RemoveVersion() []byte {
  return v[1:]
}

...
```

pkg/wallet/wallet.go
```diff
@@ -542,18 +631,21 @@ func (wallet *Wallet) GetNonce() string {
...
 
 func addressFromPublicKey(pubKeyBytes []byte) string {
-	addr := blake3.Sum256(pubKeyBytes)
-	return UserAddressPrefix + base58.CheckEncode(addr[:], Base58Version)
+	versionedPubKey := append([]byte{PubKeyVersion}, pubKeyBytes...)
```
gregLibert, line 634: You should use the helper instead of doing it manually here.

pkg/wallet/wallet.go
```diff
@@ -151,7 +163,12 @@ func (w *Wallet) Unprotect(password string) *WalletError {
 		return &WalletError{fmt.Errorf("opening the private key seal: %w", err), utils.WrongPassword}
 	}
 
-	w.KeyPair.PrivateKey = pk
+	privateKey, err := CheckPrivKeyVersion(pk)
+	if err != nil {
+		return &WalletError{fmt.Errorf("while checking private key version: %w", err), utils.ErrInvalidPrivateKey}
```
gregLibert, line 168: What is this horror ? Why do we need a Wallet Error struct ?

gregLibert (changes requested):

pkg/wallet/wallet.go
Thykof: this is the version of the wallet file format standard, not the version of a blockchain element (address, keys, signature)

pkg/wallet/wallet.go
Thykof: This function returns a WalletError to be able to specify an error and a error code. The error code is included in the response http call or to the event send to wails. This way we can return an error and the type of the error with the error code. The error code can be used to translate the error message shown to the user.

pkg/wallet/wallet.go
Thykof: done in https://github.com/massalabs/station-massa-wallet/pull/530/commits/576f72db2bd3d20acb805b4ebb49eb47692138fc

pkg/wallet/wallet.go
Thykof: done in https://github.com/massalabs/station-massa-wallet/pull/530/commits/0a5c7c0381859807eedbe1b36ccda300f64171ed

pkg/wallet/wallet.go
Thykof: see related PR: https://github.com/massalabs/massa-standards/pull/101/files

ASAPSegfault (changes requested):

internal/handler/wallet/rolls.go
```diff
@@ -66,11 +66,11 @@ func (t *tradeRolls) Handle(params operations.TradeRollsParams) middleware.Respo
 			})
 	}
 
-	operation, err := doTradeRolls(wlt, amount, fee, *params.Body.Side, t.massaClient)
-	if err != nil {
-		errStr := fmt.Sprintf("error %sing rolls coin: %v", *params.Body.Side, err.Error())
+	operation, tradeRollError := doTradeRolls(wlt, amount, fee, *params.Body.Side, t.massaClient)
+	if tradeRollError != nil {
+		errStr := fmt.Sprintf("error %sing rolls coin: %v", *params.Body.Side, tradeRollError.Err.Error())
```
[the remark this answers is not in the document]
Thykof: I think no, %s for string and ing for the verb.

pkg/wallet/wallet.go
gregLibert: :see_no_evil:

internal/handler/wallet/rolls.go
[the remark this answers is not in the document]
ASAPSegfault: Indeed

gregLibert (approved):
:see_no_evil:

PR #535, opened 2023-07-18
format generated files

Thykof:


B-Naoufal (commented):
what about those warnings :

Thykof:
yes I know about the warnings, we will fix when rework on that part

B-Naoufal (approved):

PR #537, opened 2023-07-19
standalone: use network of station if possible

Thykof:
when I use the wallet in standalone, I still want to use the network chosen by Massa Station.

This is a technical use case that could be useful.

Current behavior:
Wallet is standalone: use buildnet

This PR introduce this behavior:
Wallet always try to fetch Massa Station network endpoint, it case of failure (Massa Station isn't running), use the buildnet.

### DIFF
--- a/pkg/network/network.go
+++ b/pkg/network/network.go
@@ -2,11 +2,10 @@ package network
 
 import (
 	"encoding/json"
-	"fmt"
 	"io"
 	"net/http"
-	"os"
 
+	"github.com/labstack/gommon/log"
 	"github.com/massalabs/station-massa-hello-world/pkg/plugin"
 )
 
@@ -14,6 +13,7 @@ const (
 	massaStationNodeEndpoint = plugin.MassaStationBaseURL + "/massa/node"
 	defaultNetwork           = "buildnet"
 	defaultNodeUrl           = "https://buildnet.massa.net/api/v2"
+	fallbackMsg              = "Fallback to default network. "
 )
 
 type NetworkInfo struct {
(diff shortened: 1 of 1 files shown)


B-Naoufal (approved):

pkg/network/network.go
```diff
@@ -2,18 +2,18 @@ package network
 
 import (
 	"encoding/json"
-	"fmt"
 	"io"
 	"net/http"
-	"os"
 
+	"github.com/labstack/gommon/log"
```
gregLibert, line 8: What about using massa-station log pkg instead ?

pkg/network/network.go
```diff
@@ -24,25 +24,25 @@ type NetworkInfo struct {
...
 // retrieve network info from endpoint
 func GetNetworkInfo() (*NetworkInfo, error) {
-	if os.Getenv("STANDALONE") == "1" {
-		return &NetworkInfo{Network: defaultNetwork, URL: defaultNodeUrl}, nil
-	}
-
 	resp, err := http.Get(massaStationNodeEndpoint)
 	if err != nil {
-		return nil, fmt.Errorf("failed to make GET request: %v", err)
+		log.Warnf(fallbackMsg+"Failed to make GET request: %v", err)
```
gregLibert, line 29:
```suggestion
logFallback("GET massa station node endpoint", err)
```

With:
```golang
func logFallback(action string, err error) {
     log.Warnf("Fallback to default network: failed to %s: %v", msg, err)
 }

pkg/network/network.go
Thykof: I would require to initlialize it, so chose a logs file, I'll do it later

Thykof: https://github.com/massalabs/station-massa-wallet/issues/538

pkg/network/network.go
Thykof: I did the fix here https://github.com/massalabs/station-massa-wallet/pull/530/commits/688a2a5b0b1f1fa50fe07fbd5a711f65509ed9a2

Thykof:
QA OK

PR #549, opened 2023-07-27
547 add an endpoint get all which return all assets imported with their balance

B-Naoufal:


api/walletApi-V0.yml
```diff
@@ -601,5 +617,10 @@ definitions:
...
-
-    
\ No newline at end of file
+  AssetInfoWithBalance:
+    allOf:
+      - $ref: "#/definitions/AssetInfo"
+      - type: object
+        properties:
+          balance:
+            type: number
```
peterjah, line 626: the balance is u256, it will not fit in a number type. Return a string instead

peterjah (changes requested):

Thykof (approved):

peterjah (approved):

PR #561, opened 2023-08-01
adjust endpoint

B-Naoufal:


Thykof (changes requested):
wallet is the software (plugin), account is the keypair

B-Naoufal:
Done in https://github.com/massalabs/station-massa-wallet/pull/563

PR #578, opened 2023-08-04
542-add warning message if some accounts are corrupted

Pierregvx:


internal/handler/wallet/getall.go
```diff
@@ -41,12 +41,35 @@ func (h *walletGetAll) Handle(params operations.AccountListParams) middleware.Re
 			})
 	}
 
-	for i := 0; i < len(wallets); i++ {
-		modelWallet := createModelWallet(wallets[i])
+	for i := 0; i < len(walletsWithoutError); i++ {
+		status := "ok"
```
Thykof, line 45: we should use a constant variable, the same as the one we use when creating a new account

internal/handler/wallet/getall.go
```diff
@@ -41,12 +41,35 @@ func (h *walletGetAll) Handle(params operations.AccountListParams) middleware.Re
...
+		modelWallet := createModelWallet(walletsWithoutError[i])
 		modelWallet.CandidateBalance = models.Amount(fmt.Sprint(infos[i].CandidateBalance))
 		modelWallet.Balance = models.Amount(fmt.Sprint(infos[i].Balance))
+		modelWallet.Status = &status
 		wlts = append(wlts, &modelWallet)
 	}
+	for u := 0; u < len(walletsWithError); u++ {
+		status := "corrupted"
+		modelWalletErr := createModelWallet(walletsWithError[u])
+		modelWalletErr.Status = &status
```
Thykof, line 55: can we modify `createModelWallet` instead of setting the Status here?

pkg/wallet/wallet.go
```diff
@@ -307,16 +308,23 @@ func LoadFile(filePath string) (Wallet, *WalletError) {
...
-		return Wallet{}, errMissingFields
+		return baseAccount, errMissingFields
 	}
 
 	account, err := accountSerialized.toAccount()
 	if err != nil {
-		return Wallet{}, &WalletError{fmt.Errorf("deserializing account '%s': %w", filePath, err), utils.ErrAccountFile}
+		return baseAccount, &WalletError{fmt.Errorf("deserializing account '%s': %w", filePath, err), utils.ErrAccountFile}
 	}
+	account.Status = "ok"
```
Thykof, line 327: create a constant variable for "ok"

pkg/wallet/wallet.go
```diff
@@ -307,16 +308,23 @@ func LoadFile(filePath string) (Wallet, *WalletError) {
 	if err != nil {
 		return Wallet{}, &WalletError{fmt.Errorf("unmarshalling file '%s': %w", filePath, err), utils.ErrAccountFile}
 	}
-
+	walletHasNickname := len(accountSerialized.Nickname) > 0
+	baseAccount := Wallet{
+		Nickname: "unknown wallet",
```
Thykof, line 313: is it possible to get the wallet nickname from the filename?

web-frontend/src/pages/Index/Index.tsx
```diff
@@ -1,28 +1,45 @@
...
-  const { error, data, isLoading } = useResource<AccountObject[]>('accounts');
+  const { error, okAccounts, corruptedAccounts, isLoading } = fetchAccounts();
+  const corruptedAccountsNames = corruptedAccounts?.map(
+    (account: AccountObject) => account.nickname,
+  );
+  const accountsStringified =
+    corruptedAccountsNames
+      ?.map(
+        (account: string, index: number) =>
+          `${account}${
```
Thykof, line 21:
```suggestion
        (name: string, index: number) =>
          `${name}${
```

plugin.zip
Thykof, line 1: remove this

Thykof (changes requested):

pkg/wallet/wallet.go
Thykof: now you can use it!

pkg/wallet/wallet.go
```diff
@@ -307,16 +310,26 @@ func LoadFile(filePath string) (Wallet, *WalletError) {
 	if err != nil {
 		return Wallet{}, &WalletError{fmt.Errorf("unmarshalling file '%s': %w", filePath, err), utils.ErrAccountFile}
 	}
-
+	walletHasNickname := len(accountSerialized.Nickname) > 0
+	_, nicknameFromFileName := filepath.Split(filePath)
+	nicknameFromFileName = strings.TrimPrefix(nicknameFromFileName, "wallet_")
+	nicknameFromFileName = strings.TrimSuffix(nicknameFromFileName, ".yaml")
```
Thykof, line 316: you can extract this into a function near https://github.com/massalabs/station-massa-wallet/blob/86040bea9ca6414c2087770de7968e2b08548028/pkg/wallet/wallet.go#L389

Thykof (changes requested):

pkg/wallet/wallet.go
Thykof:
```go
```

Thykof: you can create this function and put it bellow that function: https://github.com/massalabs/station-massa-wallet/blob/86040bea9ca6414c2087770de7968e2b08548028/pkg/wallet/wallet.go#L389

```go
func NicknameFromFilePath(filePath string): string {
	_, nicknameFromFileName := filepath.Split(filePath)
	nicknameFromFileName = strings.TrimPrefix(nicknameFromFileName, "wallet_")
	nicknameFromFileName = strings.TrimSuffix(nicknameFromFileName, ".yaml")

	return nicknameFromFileName
}
```

Thykof: in golang, we usually don't prefix the getter functions with `get`

Thykof (approved):

pivilartisant (approved):

PR #603, opened 2023-08-14
602 add asset address to addasset return

B-Naoufal:


pivilartisant (approved):

api/walletApi-V0.yml
```diff
@@ -661,6 +661,8 @@ definitions:
     description: Token informations
     type: object
     properties:
+      assetAddress:
```
Thykof, line 664: why not just `address`?

internal/handler/wallet/get_all_assets.go
```diff
@@ -71,9 +71,8 @@ func (h *getAllAssets) Handle(params operations.GetAllAssetsParams) middleware.R
 
 		// Create the asset info with balance and append it to the result slice
 		assetWithBalance := &models.AssetInfoWithBalance{
-			AssetInfo:    assetInfo,
-			Balance:      balance,
-			AssetAddress: assetAddress,
```
Thykof, line 76: why?

Thykof (changes requested):

internal/handler/wallet/get_all_assets.go
B-Naoufal: now AssetInfo already contains the AssetAddress

PR #604, opened 2023-08-14
Assets Confirmation Flow

pivilartisant:
Complete asset import flow. 

Major changes : input form is now part of AssetsModalImport -> I change the level because I needed to send info to the ImportResult screen... it was so we can have less params

Screenshot 2023-08-16 at 14 18 28

PR #605, opened 2023-08-16
Fix sonar and refactor

Thykof:


web-frontend/src/const/assets/assets.ts
```diff
@@ -1,17 +1,5 @@
 export const XMA = 'XMA';
 
-export interface ITokenData {
```
pivilartisant, line 3: I believe this interface is necessary no ? because it's what the component receives

web-frontend/src/mirage/models/index.ts
```diff
@@ -3,13 +3,13 @@ import { ModelDefinition } from 'miragejs/-types';
 
 import { AccountObject } from '../../models/AccountModel';
 import { TransferModel } from '../../models/TransferModel';
-import { IToken } from '@/models/AssetModel';
+import { Token } from '@/models/AssetModel';
```
pivilartisant, line 6: replace token by "Asset" to limit conflicts with https://github.com/massalabs/station-massa-wallet/pull/604

web-frontend/src/const/assets/assets.ts
Thykof: As far as I saw it was a duplicate of IToken (now named Asset)

pivilartisant: it's a fake duplicate, it takes some elements from Asset interface and adds a bunch of things that the component receives, but it's also defined in the component, not sure why it's here twice

pivilartisant (approved):

PR #606, opened 2023-08-16
597 sign operation extract callSC operation info and send it to wails popup

B-Naoufal:


internal/handler/wallet/sign.go
```diff
@@ -42,14 +53,49 @@ func (s *walletSign) Handle(params operations.SignParams) middleware.Responder {
 		return resp
 	}
 
+	op, _ := base64.StdEncoding.DecodeString(params.Body.Operation.String())
```
Thykof, line 56: handle the error

Thykof (changes requested):

internal/handler/wallet/sign.go
```diff
@@ -53,7 +53,8 @@ func (s *walletSign) Handle(params operations.SignParams) middleware.Responder {
 		return resp
 	}
 
-	op, _ := base64.StdEncoding.DecodeString(params.Body.Operation.String())
+	op, err := base64.StdEncoding.DecodeString(params.Body.Operation.String())
+	fmt.Println("base64 decoding error: %w", err)
```
Thykof, line 57: no, here you print an error message even when there is no error. What you need to do is to check is the error is nil and if no, handle the error

Thykof (changes requested):

internal/handler/wallet/sign.go
```diff
@@ -54,7 +54,13 @@ func (s *walletSign) Handle(params operations.SignParams) middleware.Responder {
 	}
 
 	op, err := base64.StdEncoding.DecodeString(params.Body.Operation.String())
-	fmt.Println("base64 decoding error: %w", err)
+	if err != nil {
+		return operations.NewSignInternalServerError().WithPayload(
```
Thykof, line 58: this is 500, not 400

Thykof (changes requested):

internal/handler/wallet/sign.go
```diff
@@ -42,14 +53,49 @@ func (s *walletSign) Handle(params operations.SignParams) middleware.Responder {
...
+	// Just For test
+	// To Delete
+	decodedMsg, err := sendoperation.DecodeMessage64(params.Body.Operation.String())
+	if err != nil {
+		fmt.Println("Error decoding message:", err)
+	}
+
+	callSC, err := callsc.DecodeMessage(decodedMsg)
+	if err != nil {
+		// TODO : handle all cases , CallSC is well decoded , we need to do the same for ExecuteSC
```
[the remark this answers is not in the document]
Thykof: return a 400

internal/handler/wallet/sign.go
```diff
@@ -42,14 +53,55 @@ func (s *walletSign) Handle(params operations.SignParams) middleware.Responder {
...
 	var correlationId models.CorrelationID
 	if params.Body.CorrelationID != nil {
 		correlationId, resp = handleWithCorrelationId(wlt, params, s.gc)
 	} else {
 
-		promptRequest := prompt.PromptRequest{
-			Action: walletapp.Sign,
-			Msg:    fmt.Sprintf("Unprotect wallet %s", wlt.Nickname),
+		var promptRequest prompt.PromptRequest
+		if callSC != nil {
```
Thykof, line 84: this part is not maintainable.

internal/handler/wallet/sign.go
```diff
@@ -83,7 +134,7 @@ func (s *walletSign) Handle(params operations.SignParams) middleware.Responder {
 
 	signature, err := wlt.Sign(op)
 	if err != nil {
-		return operations.NewSignInternalServerError().WithPayload(
+		return operations.NewSignBadRequest().WithPayload(
```
Thykof, line 137: no, this is really a internal server error if we can't sign, not an error due to user inputs.

wails-frontend/src/pages/passwordPrompt.tsx
```diff
@@ -58,6 +67,26 @@ function TransferLayout(props: PromptRequestTransferData) {
   );
 }
 
+function SignLayout(props: PromptRequestCallSCData) {
+  const { GasLimit, Coins, Address, CalledFunction, OperationType } = props;
+
+  return (
+    <div>
+      <div>Operation Type: {OperationType}</div>
```
Thykof, line 75: `div`? why not `p`?

Thykof (changes requested):

wails-frontend/src/pages/passwordPrompt.tsx
B-Naoufal: the front end will be upgraded by the front experts , my code is just aa proof that it's working well

Thykof (approved):

PR #618, opened 2023-08-21
rename xma to mas

Thykof:


pkg/assets/asset.go
```diff
@@ -249,9 +249,9 @@ func GetAssetsJSONPath() (string, error) {
 }
 
 func XMAInfo() models.AssetInfo {
```
thomas-senechal, line 251: Shouldn't this function be also renamed ? 

```suggestion
func MASInfo() models.AssetInfo {
```

thomas-senechal (approved):

PR #626, opened 2023-08-22
Check manifest version in the release ci

Thykof:


.github/workflows/build.yml
```diff
@@ -77,6 +77,15 @@ jobs:
             ${{ env.TARGET_NAME }}${{ matrix.ext }}
             wallet.svg
             manifest.json
+      - name: Check the manifest version
+        if: ${{ inputs.tag_name != '' }}
+        run: |
+          sudo apt-get install -y jq
+          version=$(jq -r '.key' manifest.json)
```
thomas-senechal, line 84: Isn't it supposed to be `version` ? 🤔 

```suggestion
          version=$(jq -r '.version' manifest.json)
```

thomas-senechal: Based on this:
https://stackoverflow.com/questions/61919141/read-json-file-in-github-actions#comment126579492_61919791

thomas-senechal:
> we need to find a way to test this before merging

I found a simple way to test it manually:
I forked the repository and merged the branch into main, then, I made a release with a version number that wasn't equal to the one in the Manifest. 
The CI failed thanks to the error message: https://github.com/thomas-senechal/station-massa-wallet/actions/runs/5960583935

I think we should move the version check before creating the matrix, because right now, we check in the matrix, which is useless since if one matrix fails, the others will fail too. Would be faster and if one day we need to pay for runners, will cost less

peterjah (approved):

PR #672, opened 2023-09-06
608 send ico quest results from massa wallet

thomas-senechal:


internal/handler/wallet/transfer.go
```diff
@@ -94,6 +95,10 @@ func (t *transferCoin) Handle(params operations.TransferCoinParams) middleware.R
 
 	t.prompterApp.EmitEvent(walletapp.PromptResultEvent,
 		walletapp.EventData{Success: true, CodeMessage: utils.MsgTransferSuccess})
+
+	//ICOQUEST: To be removed when ICO is over
+	//nolint:errcheck
+	ico.ValidateQuest("PAY_INVOICE", *params.Body.RecipientAddress)
```
peterjah, line 101: how to you know its a transfert from an Invoice? and not just a simple transfer ?
Not sure how to do this but maybe you can only do it frontend side

peterjah (changes requested):

Thykof (approved):

internal/handler/wallet/transfer.go
thomas-senechal: I had a quick discussion with Ena to check what we need exactly, and you are right, we want to check that the user used the generated link

So, I will store the information somewhere (localstorage or something like that) when we are redirected from https://github.com/massalabs/station-massa-wallet/blob/main/web-frontend/src/pages/TransferCoins/SendCoins/SendRedirect.tsx 
Then, when the API call of the send coins succeeded, I will check if the receiving address matches the one that has been stored

Thykof (approved):

peterjah (approved):

PR #674, opened 2023-09-06
macos: add logo in dock

Thykof:


.github/workflows/build.yml
```diff
@@ -2,7 +2,7 @@ name: Build
 
 on:
   push:
-    branches: [main]
+    # branches: [main]
```
peterjah, line 5: to uncomment !

Taskfile.yml
```diff
@@ -4,12 +4,19 @@ tasks:
...
-      - cmd: sudo apt install -y libgtk-3-dev libwebkit2gtk-4.0-dev
+      - cmd: sudo apt update && sudo apt install -y libgtk-3-0 libwebkit2gtk-4.0-dev gcc-aarch64-linux-gnu
         platforms: [linux]
+        ignore_error: true
+      - cmd: brew install mitchellh/gon/gon
+        platforms: [darwin]
+        ignore_error: true
       - cmd: go install github.com/go-swagger/go-swagger/cmd/swagger@latest
-      - cmd: go install github.com/wailsapp/wails/v2/cmd/wails@v2.5.1
+      - cmd: go install github.com/wailsapp/wails/v2/cmd/wails@v2.6.0
```
peterjah, line 14: wails is still in version 2.5.1 in go.sum, so you should update both sides

peterjah (changes requested):

.github/workflows/build.yml
Thykof: I think we can keep the build for any push on any branch, like in Station

pkg/wallet/wallet.go
```diff
@@ -226,13 +227,18 @@ func GetWorkDir() (string, error) {
 		return "", fmt.Errorf("getting executable path: %w", err)
 	}
 
+	if runtime.GOOS == "darwin" {
+		return filepath.Dir(filepath.Dir(filepath.Dir(filepath.Dir(ex)))), nil
```
thomas-senechal, line 231: What is supposed to be this directory ? 
I think we should add a comment ^^

Taskfile.yml
```diff
@@ -52,21 +59,20 @@ tasks:
   build-linux:
     platforms: [linux]
     cmds:
-      - cmd: mkdir -p build/wallet-plugin
```
thomas-senechal, line 55: Are you sure the `build` directory will be always created ? (same for MacOS and Windows)

thomas-senechal (approved):

Taskfile.yml
Thykof: yes because it's on github and -p option will create the full path, including build directory

PR #703, opened 2023-09-11
fix base app when switch

mazmassa:


longgamma-massa:
We have an issue:
Now we get an error each time I try to switch wallet accounts! 

Check video for details

https://github.com/massalabs/station-massa-wallet/assets/113043914/8e2ba871-a069-4931-ab05-f65fc73816ed

PR #704, opened 2023-09-12
Set Wails Always on Top

pivilartisant:
before : wails popup behaves like normal window 
after : wails popup will be always on top, user can minimize the window but it will not stack under if the user changes focus

### DIFF
--- a/pkg/wails/wails.go
+++ b/pkg/wails/wails.go
@@ -16,6 +16,7 @@ func NewWailsApp(app *wApp.WalletApp, assets embed.FS) *application.Application 
 		Width:       513,
 		Height:      440,
 		StartHidden: true,
+		AlwaysOnTop: true,
 		AssetServer: &assetserver.Options{
 			Assets: assets,
 		},


pivilartisant:
Since it's part of a TBD milestone we might not need to merge right now

mazmassa (approved):

Thykof:
@pivilartisant could you describe before/after behavior in the PR description?

thomas-senechal:
Did you had the bug mentionned in the issue ?
Did you discussed with someone about using the `AlwaysOnTop` option as the solution ?

In my opinion, it's a bit overkill because, if I'm not wrong, that means that the window can't be "minimized" or behind another one, which can be annoying especially on small screens

pivilartisant:
> Did you had the bug mentionned in the issue ? Did you discussed with someone about using the `AlwaysOnTop` option as the solution ?
> 
> In my opinion, it's a bit overkill because, if I'm not wrong, that means that the window can't be "minimized" or behind another one, which can be annoying especially on small screens

Window can be minimized, but indeed it will always be on top during any other interaction

pivilartisant:
> > Did you had the bug mentionned in the issue ? Did you discussed with someone about using the `AlwaysOnTop` option as the solution ?
> > In my opinion, it's a bit overkill because, if I'm not wrong, that means that the window can't be "minimized" or behind another one, which can be annoying especially on small screens
> 
Do I didn't have the bug mentionned in the issue 
> Window can be minimized, but indeed it will always be on top during any other interaction

pivilartisant:
> Did you had the bug mentionned in the issue ? Did you discussed with someone about using the `AlwaysOnTop` option as the solution ?
> 
> In my opinion, it's a bit overkill because, if I'm not wrong, that means that the window can't be "minimized" or behind another one, which can be annoying especially on small screens

No I did not have the bug mentioned in the issue unfortunately 
Window can be minimized, but indeed it will always be on top during any other interaction

pivilartisant:
We chose to not go forward with this pr as it did not improve UX

PR #708, opened 2023-09-12
amount-of-coins-sent-in-a-callsc-displayed-is-wrong

B-Naoufal:


Thykof (approved):

wails-frontend/src/pages/PasswordPromptHandler/SignSC/CallSc.tsx
```diff
@@ -22,10 +22,13 @@ export function CallSc(props: PromptRequestData) {
     Description,
   } = props;
 
-  const toAddInHeigthDescription = Description ? 200 : 0;
+  // Convert nano mass to mass
+  const coinsInMass = coinsInNanoMass / 1000000000;
```
Thykof, line 26: can you see if it's possible to use the massa-web3 convert function?

Thykof: https://github.com/massalabs/massa-web3/blob/815357a17b0b44be8759a4abcc643291f7e57c4f/src/utils/converters.ts#L50

peterjah (changes requested):
This whole thing must be refactored using string/bigint instead of number for every "amounts"

B-Naoufal:
will be covered here : https://github.com/massalabs/station-massa-wallet/pull/716

PR #719, opened 2023-09-14
Fix Exit Account Creation Logic

pivilartisant:
this adds an additional check to verify that we are not in the /index root which was the route used after deleting an account. 

The flow is intact during account creation and adding an account with the dropdown


mazmassa (approved):

Thykof:
We will merge this one after release v0.2.6

PR #734, opened 2023-09-15
Add Jest to Wails

pivilartisant:
We do not have a clear strategy with unit testing and what technos we are going to use i will keep this in draft until decisions on the subject have been made


B-Naoufal (approved):

wails-frontend/src/__ tests __/massaFormat.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+  // Format the number as a string with separators
+  const formattedNum = formatValue({
+    value: numInMas,
+    groupSeparator: ',',
+    decimalSeparator: '.',
+    decimalScale: 9,
+  });
+  return removeTrailingZeros(formattedNum);
+}
```
peterjah, line 33: why redefining this?  you are not testing the actual code :/

pivilartisant: massFormat.tsx seems to host some code that is used else where, when i run the test cmd, it fails beacause jest can't find in correct module (which is generated by go in the ./wailsjs dir)

pivilartisant: Screenshot 2023-09-18 at 14 20 52

peterjah (approved):

PR #750, opened 2023-09-20
Fix crash at account dir creation

thomas-senechal:


gregLibert (approved):

Thykof (approved):
nice, do you understand why `os.IsNotExist(err)` returns true while the folder actually exists?

Thykof:
> nice, do you understand why `os.IsNotExist(err)` returns true while the folder actually exists?

maybe concurrency

PR #753, opened 2023-09-20
logs in a file

Thykof:


Thykof:
Work on Ubuntu and MacOS but not on windows:

```
[wallet-plugin] Error: panic: [wallet-plugin] Error: runtime error: invalid memory address or nil pointer dereference
[signal 0xc0000005 code=0x0 addr=0x0 pc=0xcab62d]

goroutine 20 [running]:
[wallet-plugin] Error: go.uber.org/zap.(*SugaredLogger).log(0x0, 0x0, {0xf09b19, 0x21}, {0xc000b6aa80, 0x1, 0x1}, {0x0, 0x0, 0x0})
        C:/Users/Nathan/go/pkg/mod/go.uber.org/zap@v1.24.0/sugar.go:295 +0xed
go.uber.org/zap.(*SugaredLogger).Infof(0x0, {0xf09b19, 0x21}, {0xc000b6aa80, 0x1, 0x1})
        C:/Users/Nathan/go/pkg/mod/go.uber.org/zap@v1.24.0/sugar.go:172 +0x125
github.com/massalabs/station/pkg/logger.Infof({0xf09b19, 0x21}, {0xc000b6aa80, 0x1, 0x1})
```

it seems like `s` in https://github.com/uber-go/zap/blob/master/sugar.go#L134C33-L134C33 is nil

peterjah (approved):

PR #754, opened 2023-09-21
740-1 add objects

gregLibert:


pkg/types/object/object.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	return err
+}
+
+// MustBytes returns the binary representation of the object.
+func (o *Object) MustBytes() []byte {
+	data, err := o.MarshalBinary()
+	if err != nil {
+		panic(err)
```
Thykof, line 113: are you sure we can afford to panic?

gregLibert: Yes, because:
- nobody is forced to use this function, you can use marshalBinary directly if you want to handle the error yourself
- there is no error returned by marshalBinary today
- having to handle an error in the code knowing that there is no udnerlying error is a to masochist IMHO

peterjah (approved):
Nice!
If we follow the logic of removing sensitive clear data from the memory, should't we use the same process for the password?

Thykof:
> Nice! If we follow the logic of removing sensitive clear data from the memory, should't we use the same process for the password?

yes! we will. If you think in that PR something is missing, leave a review ;)

pkg/types/object/object_test.go
```diff
@@ -0,0 +1,94 @@
+// Package object provides functionalities for handling different types of Massa objects.
+// It supports encoding and decoding operations with appropriate prefixes and versioning.
```
thomas-senechal, line 2: I don't think it's relevant since we already gave it in the object.go file
```suggestion
```

thomas-senechal (approved):
Looks good to me ! 

Just one non blocking comment

PR #760, opened 2023-09-26
740-2 account package

Thykof:
- [ ] handle version missing in the yaml

### DIFF
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/massalabs/station-massa-wallet/pkg/utils"
 	"github.com/massalabs/station-massa-wallet/pkg/wallet"
+	"github.com/massalabs/station-massa-wallet/pkg/wallet/account"
 	"github.com/wailsapp/wails/v2/pkg/runtime"
 )
 
@@ -135,5 +136,5 @@ func (a *WalletApp) IsNicknameUnique(nickname string) bool {
 }
 
 func (a *WalletApp) IsNicknameValid(nickname string) bool {
-	return wallet.NicknameIsValid(nickname)
+	return account.NicknameIsValid(nickname)
 }
--- a/pkg/app/app.go
+++ b/pkg/app/app.go
@@ -39,7 +39,9 @@ func NewWalletApp() *WalletApp {
 		Shutdown:    false,
 		IsListening: false,
 	}
+
 	go app.cleanExit()
+
 	return app
 }
 
@@ -55,6 +57,7 @@ func (a *WalletApp) Startup(ctx context.Context) {
 
 func (a *WalletApp) BeforeClose(ctx context.Context) bool {
 	a.Hide()
+
 	if a.Shutdown {
 		return false
 	}
@@ -64,6 +67,7 @@ func (a *WalletApp) BeforeClose(ctx context.Context) bool {
 		fmt.Println("canceling prompt before closing")
 		a.CtrlChan <- Cancel
 	}
+
 	runtime.WindowReloadApp(a.Ctx)
 
 	return true
--- a/pkg/prompt/import.go
+++ b/pkg/prompt/import.go
@@ -7,6 +7,7 @@ import (
 	walletapp "github.com/massalabs/station-massa-wallet/pkg/app"
 	"github.com/massalabs/station-massa-wallet/pkg/utils"
 	"github.com/massalabs/station-massa-wallet/pkg/wallet"
+	"github.com/massalabs/station-massa-wallet/pkg/wallet/account"
 )
 
 func handleImportPrompt(prompterApp WalletPrompterInterface, input interface{}) (*wallet.Wallet, bool, error) {
@@ -29,7 +30,7 @@ func handleImportFile(prompterApp WalletPrompterInterface, filePath string) (*wa
 			walletapp.EventData{Success: false, CodeMessage: utils.ErrInvalidFileExtension})
 		return nil, false, fmt.Errorf(InvalidAccountFileErr)
 	}
-	account, loadErr := wallet.LoadFile(filePath)
+	acc, loadErr := wallet.LoadFile(filePath)
 	if loadErr != nil {
 		errStr := fmt.Sprintf("%v: %v", AccountLoadErr, loadErr.Err)
 		prompterApp.EmitEvent(walletapp.PromptResultEvent,
@@ -38,16 +39,16 @@ func handleImportFile(prompterApp WalletPrompterInterface, filePath string) (*wa
 	}
 
 	// Validate nickname
-	if !wallet.NicknameIsValid(account.Nickname) {
+	if !account.NicknameIsValid(acc.Nickname) {
 		errorCode := utils.ErrInvalidNickname
-		fmt.Printf("error while importing: invalid nickname: '%s'\n", account.Nickname)
+		fmt.Printf("error while importing: invalid nickname: '%s'\n", acc.Nickname)
 		prompterApp.EmitEvent(walletapp.PromptResultEvent,
 			walletapp.EventData{Success: false, CodeMessage: errorCode})
 		return nil, false, fmt.Errorf(errorCode)
 	}
 
 	// Validate nickname uniqueness
-	err := wallet.NicknameIsUnique(account.Nickname)
+	err := wallet.NicknameIsUnique(acc.Nickname)
 	if err != nil {
 		errorCode := utils.ErrDuplicateNickname
 		prompterApp.EmitEvent(walletapp.PromptResultEvent,
@@ -56,7 +57,7 @@ func handleImportFile(prompterApp WalletPrompterInterface, filePath string) (*wa
 	}
 
 	// Validate unique private key
-	err = wallet.AddressIsUnique(account.Address)
+	err = wallet.AddressIsUnique(acc.Address)
 	if err != nil {
 		errorCode := utils.ErrDuplicateKey
 		prompterApp.EmitEvent(walletapp.PromptResultEvent,
@@ -64,7 +65,7 @@ func handleImportFile(prompterApp WalletPrompterInterface, filePath string) (*wa
 		return nil, false, fmt.Errorf(errorCode)
 	}
 
-	return &account, false, nil
+	return &acc, false, nil
 }
 
 func handleImportPrivateKey(prompterApp WalletPrompterInterface, walletInfo walletapp.ImportFromPKey) (*wallet.Wallet, bool, error) {
--- a/pkg/prompt/import.go
+++ b/pkg/prompt/import.go
@@ -30,20 +30,24 @@ func handleImportFile(prompterApp WalletPrompterInterface, filePath string) (*wa
 			walletapp.EventData{Success: false, CodeMessage: utils.ErrInvalidFileExtension})
 		return nil, false, fmt.Errorf(InvalidAccountFileErr)
 	}
+
 	acc, loadErr := wallet.LoadFile(filePath)
 	if loadErr != nil {
 		errStr := fmt.Sprintf("%v: %v", AccountLoadErr, loadErr.Err)
 		prompterApp.EmitEvent(walletapp.PromptResultEvent,
 			walletapp.EventData{Success: false, CodeMessage: loadErr.CodeErr})
+
 		return nil, false, fmt.Errorf(errStr)
 	}
 
 	// Validate nickname
 	if !account.NicknameIsValid(acc.Nickname) {
 		errorCode := utils.ErrInvalidNickname
+
 		fmt.Printf("error while importing: invalid nickname: '%s'\n", acc.Nickname)
 		prompterApp.EmitEvent(walletapp.PromptResultEvent,
 			walletapp.EventData{Success: false, CodeMessage: errorCode})
+
 		return nil, false, fmt.Errorf(errorCode)
 	}
 
@@ -53,6 +57,7 @@ func handleImportFile(prompterApp WalletPrompterInterface, filePath string) (*wa
 		errorCode := utils.ErrDuplicateNickname
 		prompterApp.EmitEvent(walletapp.PromptResultEvent,
 			walletapp.EventData{Success: false, CodeMessage: errorCode})
+
 		return nil, false, fmt.Errorf(errorCode)
 	}
 
@@ -62,6 +67,7 @@ func handleImportFile(prompterApp WalletPrompterInterface, filePath string) (*wa
 		errorCode := utils.ErrDuplicateKey
 		prompterApp.EmitEvent(walletapp.PromptResultEvent,
 			walletapp.EventData{Success: false, CodeMessage: errorCode})
+
 		return nil, false, fmt.Errorf(errorCode)
 	}
 
@@ -74,6 +80,7 @@ func handleImportPrivateKey(prompterApp WalletPrompterInterface, walletInfo wall
 		errStr := fmt.Sprintf("%v: %v", ImportPrivateKeyErr, importErr.Err)
 		prompterApp.EmitEvent(walletapp.PromptResultEvent,
 			walletapp.EventData{Success: false, CodeMessage: importErr.CodeErr})
+
 		return nil, false, fmt.Errorf(errStr)
 	}
 
--- a/pkg/types/encrypted_private_key.go
+++ b/pkg/types/encrypted_private_key.go
@@ -141,8 +141,8 @@ func (e *EncryptedPrivateKey) PrivateKeyTextInClear(password *memguard.LockedBuf
 	return privateKeyBuffer, nil
 }
 
-// PasswordIsValid returns true if the password is valid for the account. It destroys the password.
-func (e *EncryptedPrivateKey) PasswordIsValid(password *memguard.LockedBuffer, salt, nonce, encryptedKey []byte) bool {
+// HasAccess returns true if the password is valid for the account. It destroys the password.
+func (e *EncryptedPrivateKey) HasAccess(password *memguard.LockedBuffer, salt, nonce, encryptedKey []byte) bool {
 	privateKeyInClear, err := privateKey(password, salt, nonce, e.Data)
 	if err != nil {
 		return false
--- a/pkg/types/encrypted_private_key.go
+++ b/pkg/types/encrypted_private_key.go
@@ -93,7 +93,7 @@ func (e *EncryptedPrivateKey) Sign(password *memguard.LockedBuffer, salt, nonce,
 
 	privateKeyInClear, err := privateKey(password, salt, nonce, e.Data)
 	if err != nil {
-		return nil, fmt.Errorf("Sign: %w", err)
+		return nil, fmt.Errorf("failed to get private key: %w", err)
 	}
 
 	defer privateKeyInClear.Destroy()
--- a/pkg/types/encrypted_private_key.go
+++ b/pkg/types/encrypted_private_key.go
@@ -50,6 +50,7 @@ func (e *EncryptedPrivateKey) MarshalText() ([]byte, error) {
 	if err := e.validate(); err != nil {
 		return nil, err
 	}
+
 	return e.Object.MarshalText()
 }
 
@@ -145,6 +146,7 @@ func (e *EncryptedPrivateKey) PasswordIsValid(password *memguard.LockedBuffer, s
 	if err != nil {
 		return false
 	}
+
 	privateKeyInClear.Destroy()
 
 	return err == nil
--- a/pkg/types/encrypted_private_key.go
+++ b/pkg/types/encrypted_private_key.go
@@ -106,7 +106,7 @@ func (e *EncryptedPrivateKey) Sign(password *memguard.LockedBuffer, salt, nonce,
 func (e *EncryptedPrivateKey) PublicKey(password *memguard.LockedBuffer, salt, nonce []byte) (*PublicKey, error) {
 	privateKeyInClear, err := privateKey(password, salt, nonce, e.Data)
 	if err != nil {
-		return nil, fmt.Errorf("PublicKey: %w", err)
+		return nil, fmt.Errorf("failed to get private key: %w", err)
 	}
 
 	publicKeyBytes := ed25519.PrivateKey(privateKeyInClear.Bytes()).Public().(ed25519.PublicKey)
@@ -126,7 +126,7 @@ func (e *EncryptedPrivateKey) PublicKey(password *memguard.LockedBuffer, salt, n
 func (e *EncryptedPrivateKey) PrivateKeyTextInClear(password *memguard.LockedBuffer, salt, nonce, encryptedKey []byte) (*memguard.LockedBuffer, error) {
 	privateKeyInClear, err := privateKey(password, salt, nonce, encryptedKey)
 	if err != nil {
-		return nil, fmt.Errorf("PrivateKeyTextInClear: %w", err)
+		return nil, fmt.Errorf("failed to get private key: %w", err)
 	}
 
 	seed := ed25519.PrivateKey(privateKeyInClear.Bytes()).Seed()
@@ -135,6 +135,7 @@ func (e *EncryptedPrivateKey) PrivateKeyTextInClear(password *memguard.LockedBuf
 	seedBuffer := memguard.NewBufferFromBytes(seed)
 
 	privateKey := e.Kind.Prefix() + base58.CheckEncode(seedBuffer.Bytes(), e.Version)
+	seedBuffer.Destroy()
 	privateKeyBuffer := memguard.NewBufferFromBytes([]byte(privateKey))
 
 	return privateKeyBuffer, nil
@@ -149,7 +150,7 @@ func (e *EncryptedPrivateKey) PasswordIsValid(password *memguard.LockedBuffer, s
 
 	privateKeyInClear.Destroy()
 
-	return err == nil
+	return true
 }
 
 // privateKey returns the private key in clear. Password is destroyed.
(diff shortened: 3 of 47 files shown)


pkg/wallet/account/account.go
```diff
@@ -0,0 +1,152 @@
...
+		Nickname:     nickname,
+		Address:      *address,
+		Salt:         salt,
+		Nonce:        nonce,
+		CipheredData: cipheredData,
+		PublicKey:    publicKey,
+	}, nil
+}
+
+func NewFromPrivateKey(password *memguard.LockedBuffer, nickname string, privateKeyText *memguard.LockedBuffer) (*Account, error) {
```
Thykof, line 89: there is probably a way to avoid code duplication between those to constructor functions

pkg/app/app.go
```diff
@@ -90,50 +88,8 @@ func (a *WalletApp) Hide() {
...
-	}
-
-	if filePath == "" {
-		return selectFileResult{Err: utils.ErrNoFile, CodeMessage: utils.ErrNoFile}
-	}
-
-	wallet, loadErr := wallet.LoadFile(filePath)
-	if loadErr != nil {
-		fmt.Printf("error while loading file: %v", loadErr.Err)
-		return selectFileResult{Err: loadErr.Err.Error(), CodeMessage: loadErr.CodeErr}
```
peterjah, line 117: why deleting all theses methods ?? they are binded function that are called by the frontend

Thykof: you're right

pkg/prompt/import.go
```diff
@@ -38,16 +39,16 @@ func handleImportFile(prompterApp WalletPrompterInterface, filePath string) (*wa
 	}
 
 	// Validate nickname
-	if !wallet.NicknameIsValid(account.Nickname) {
+	if !account.NicknameIsValid(acc.Nickname) {
```
gregLibert, line 42: What about providing a wallet.CheckAccount(acc) function that will perform all checks ?

pkg/types/encrypted_private_key.go
```diff
@@ -78,11 +79,11 @@ func (a *EncryptedPrivateKey) UnmarshalBinary(data []byte) error {
 }
 
 // Sign signs the given data using the private key.
-// password is destroyed just after being used even if an error occurs.
-func (e *EncryptedPrivateKey) Sign(password *memguard.LockedBuffer, salt, nonce, data []byte) ([]byte, error) {
+// Guarded password is destroyed.
+func (e *EncryptedPrivateKey) Sign(guardedPassword *memguard.LockedBuffer, salt, nonce, data []byte) ([]byte, error) {
```
gregLibert, line 83: Why did you rename *password* to *guardedPassword*? The code readability has been made heavier without the information provided being important for the reader, because anyway you added an implementation comment explaining that the password will be destroyed.

pkg/types/encrypted_private_key.go
```diff
@@ -92,9 +93,9 @@ func (e *EncryptedPrivateKey) Sign(password *memguard.LockedBuffer, salt, nonce,
 	return append([]byte{EncryptedPrivateKeyLastVersion}, ed25519.Sign(privateKeyInClear.Bytes(), digest[:])...), nil
 }
 
-// PublicKey returns the public key corresponding to the private key.
-func (e *EncryptedPrivateKey) PublicKey(password *memguard.LockedBuffer, salt, nonce []byte) (*PublicKey, error) {
-	privateKeyInClear, err := PrivateKey(password, salt, nonce, e.Data)
+// PublicKey returns the public key corresponding to the private key. Guarded password is destroyed.
+func (e *EncryptedPrivateKey) PublicKey(guardedPassword *memguard.LockedBuffer, salt, nonce []byte) (*PublicKey, error) {
```
gregLibert, line 97: The same goes here.

pkg/types/encrypted_private_key.go
```diff
@@ -112,9 +113,40 @@ func (e *EncryptedPrivateKey) PublicKey(password *memguard.LockedBuffer, salt, n
 	}, nil
 }
 
-func PrivateKey(password *memguard.LockedBuffer, salt, nonce, encryptedKey []byte) (*memguard.LockedBuffer, error) {
-	aeadCipher, secretKey, err := crypto.NewSecretCipher(password.Bytes(), salt[:])
-	defer password.Destroy()
+// PrivateKeyTextInClear returns the private key in clear. Guarded password is destroyed.
+func (e *EncryptedPrivateKey) PrivateKeyTextInClear(guardedPassword *memguard.LockedBuffer, salt, nonce, encryptedKey []byte) (*memguard.LockedBuffer, error) {
```
gregLibert, line 117: It's the same here.

pkg/types/encrypted_private_key.go
```diff
@@ -112,9 +113,40 @@ func (e *EncryptedPrivateKey) PublicKey(password *memguard.LockedBuffer, salt, n
...
+	privateKeyInClear, err := PrivateKey(guardedPassword, salt, nonce, encryptedKey)
+	if err != nil {
+		return nil, err
+	}
+
+	seed := ed25519.PrivateKey(privateKeyInClear.Bytes()).Seed()
```
gregLibert, line 123: privateKeyInClear.Bytes() has the same value as seed.Bytes(), why do you need to duplicate it?

pkg/types/encrypted_private_key.go
gregLibert, line 117: Why didn't you create a privateKey object with marshal and unmarhsal function as the other objects ?

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,265 @@
...
+	"github.com/massalabs/station-massa-wallet/pkg/types/object"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	AccountLastVersion = 1
+)
+
+type Account struct {
+	Version      *uint8                    `yaml:"Version"`
```
gregLibert, line 21: Why is it a pointer ?

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func New(version *uint8, nickname string, address types.Address, salt [16]byte, nonce [12]byte, cipheredData types.EncryptedPrivateKey, publicKey types.PublicKey) (*Account, error) {
+	if version == nil {
+		version = new(uint8)
+		*version = AccountLastVersion
+	}
+
+	if *version > AccountLastVersion {
+		return nil, fmt.Errorf("unsupported account version: %d", *version)
```
gregLibert, line 37: you should wrap an invalidArgument sentinel error.

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,265 @@
...
+		return nil, fmt.Errorf("validating address: %w", err)
+	}
+
+	if err := cipheredData.Validate(cipheredData.Version, object.EncryptedPrivateKey); err != nil {
+		return nil, fmt.Errorf("validating encrypted private key: %w", err)
+	}
+
+	if err := publicKey.Validate(publicKey.Version, object.PublicKey); err != nil {
+		fmt.Println("52", publicKey.Version)
+		fmt.Println("53", publicKey.Kind)
```
gregLibert, line 54: Debug code to be removed ?

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,265 @@
...
+		Nickname:     nickname,
+		Address:      address,
+		Salt:         salt,
+		Nonce:        nonce,
+		CipheredData: cipheredData,
+		PublicKey:    publicKey,
+	}, nil
+}
+
+func NewEmpty() Account {
```
gregLibert, line 69: Why do you need this function?

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,265 @@
...
+				Kind:    object.PublicKey,
+				Version: 0x00,
+				Data:    nil,
+			},
+		},
+	}
+}
+
+// NewGenerated generates a new account with a random private key. It destroys the guarded password.
+func NewGenerated(guardedPassword *memguard.LockedBuffer, nickname string) (*Account, error) {
```
gregLibert, line 100: Should be named Generate

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,265 @@
...
+	version := uint8(AccountLastVersion)
+
+	publicKeyBytes, privateKeyBytes, err := ed25519.GenerateKey(nil)
+	if err != nil {
+		return nil, fmt.Errorf("generating ed25519 key-pair: %w", err)
```
gregLibert, line 105: You should also wrap a sentinel error.

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,265 @@
...
+			},
+		},
+	}
+}
+
+// NewGenerated generates a new account with a random private key. It destroys the guarded password.
+func NewGenerated(guardedPassword *memguard.LockedBuffer, nickname string) (*Account, error) {
+	version := uint8(AccountLastVersion)
+
+	publicKeyBytes, privateKeyBytes, err := ed25519.GenerateKey(nil)
```
gregLibert, line 103: You should add a security comment explaining that the entropy is read from [crypto/rand.Reader](https://pkg.go.dev/crypto/rand#Reader) or even better, diretcly use rand as you already import it for the salt and the nonce

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,265 @@
...
+	}
+
+	aeadCipher, secretKey, err := crypto.NewSecretCipher(guardedPassword.Bytes(), salt[:])
+	guardedPassword.Destroy()
+	if err != nil {
+		return nil, fmt.Errorf("creating secret cipher: %w", err)
+	}
+
+	// secretBuffer is the secret key in clear without the version.
+	secretBuffer := memguard.NewBufferFromBytes(privateKeyBytes.Seed())
```
gregLibert, line 127: You're mixing secret key and provate key. here you should talk about the private keyu instead.

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,265 @@
...
+	Version      *uint8                    `yaml:"Version"`
+	Nickname     string                    `yaml:"Nickname,omitempty"`
+	Address      types.Address             `yaml:"Address,omitempty"`
+	Salt         [16]byte                  `yaml:"Salt,flow"`
+	Nonce        [12]byte                  `yaml:"Nonce,flow"`
+	CipheredData types.EncryptedPrivateKey `yaml:"CipheredData,flow"`
+	PublicKey    types.PublicKey           `yaml:"PublicKey,flow"`
+}
+
+func New(version *uint8, nickname string, address types.Address, salt [16]byte, nonce [12]byte, cipheredData types.EncryptedPrivateKey, publicKey types.PublicKey) (*Account, error) {
```
gregLibert, line 30: Why address, types.EncryptedPrivateKey, publicKey are not pass by refenrece instead of duplicating them ?

Furthermore cipheredData is obscure, you should use encryptedPrivateKey as you are using address or publickey.

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,265 @@
...
+	if err != nil {
+		seedBuffer.Destroy()
+		return nil, fmt.Errorf("decoding base58 private key: %w", err)
+	}
+
+	privateKey := memguard.NewBufferFromBytes(ed25519.NewKeyFromSeed(seedBuffer.Bytes()))
+	seedBuffer.Destroy()
+
+	encryptedSecret := crypto.SealSecret(aeadCipher, nonce[:], privateKey)
+	secretKey.Destroy()
```
gregLibert, line 184: I have the feeling that the same code (sealing a private key) is called multiple times. Can you create a hekper to do that only at one place.

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,265 @@
...
+		return fmt.Errorf("missing ciphered data")
+	}
+
+	if len(a.PublicKey.Data) == 0 {
+		return fmt.Errorf("missing public key")
+	}
+
+	// if a.Version == nil {
+	// 	return fmt.Errorf("missing version")
+	// }
```
gregLibert, line 258: dead code to be removed ?

gregLibert (changes requested):

pkg/wallet/account/account.go
Thykof: idk. I need help on the version

pkg/prompt/import.go
Thykof: good idea, I'll try to do that in the following PR on wallet package

pkg/wallet/account/account.go
Thykof: what about `object.ErrUnsupportedVersion`?

pkg/wallet/account/account.go
Thykof: to do this:

```go
acc := account.NewEmpty()
err = acc.Unmarshal(data)
if err != nil {
  return nil, fmt.Errorf("%w: %w", ErrUnmarshalAccount, err)
}
```

pkg/wallet/account/account.go
Thykof: I need it to be a pointer because it's a mandatory field in the yaml, and the yaml unmarshal will put value nil and so that I can throw an error if the value is nil after unmarshaling the yaml

pkg/wallet/account/account.go
gregLibert: IMHO this need is the result of the Struct  Account which is haphazard: optional objects and consequential objects (to reduce memory footprint and duplication) should be pointers, whereas objects with default values should be allocations.

pkg/wallet/account/account.go
gregLibert: I was more thinking of:
```golang
var ErrInvalidParameter = errors.New("invalid parameter")
...
if version > AccountLastVersion || version == UnknownVersion {
   return nil, return fmt.Errorf("%w: invalid version", ErrInvalidParameter)
}
```
This way, on the caller side you can easily detect that an error is about an invalid parameter, to return an http error code 422 for instance, by doing:
```golang
err := account.New(...)
if errors.Is(err, ErrInvalidParameter) {
    // handle the invalid parameter error
}
```

pkg/wallet/account/account.go
gregLibert: Your reason seems shaky: using uint8(0) as UnknownVersion value would avoid using a pointer and will be consistent with the default value of an uint8.

pkg/types/encrypted_private_key.go
gregLibert: It's weird, looking at the code of the Seed function, it should return the private key bytes: https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/crypto/ed25519/ed25519.go;l=75

pkg/wallet/account/account.go
Thykof: yes but value 0 for a version is a correct value, our standard wallet file format first version is 0.

Thykof: let's take 0 as unknown version

`UnknownVersion := uint8(0)`

pkg/wallet/account/account.go
Thykof: no need this function

acc := Account{}

peterjah (approved):

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,246 @@
...
+	"github.com/btcsuite/btcutil/base58"
+	"github.com/massalabs/station-massa-wallet/pkg/crypto"
+	"github.com/massalabs/station-massa-wallet/pkg/types"
+	"github.com/massalabs/station-massa-wallet/pkg/types/object"
+	"gopkg.in/yaml.v2"
+)
+
+const (
+	AccountLastVersion    = 1
+	AccountUnknownVersion = 0
```
gregLibert, line 19: Account prefix is useless as you're already in an Account package.

pkg/types/encrypted_private_key.go
```diff
@@ -29,74 +31,81 @@ func (a *EncryptedPrivateKey) validate() error {
...
-// password is destroyed just after being used even if an error occurs.
+// Sign signs the given data using the private key. Password is destroyed.
 func (e *EncryptedPrivateKey) Sign(password *memguard.LockedBuffer, salt, nonce, data []byte) ([]byte, error) {
 	digest := blake3.Sum256(data)
 
-	privateKeyInClear, err := PrivateKey(password, salt, nonce, e.Data)
+	privateKeyInClear, err := privateKey(password, salt, nonce, e.Data)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Sign: %w", err)
```
gregLibert, line 96: Error message should start with capital letters in go. Furthermore your error is not giving context, it simply contains the name of the function.

my suggestion:
```suggestion
		return nil, fmt.Errorf("failed to get private key: %w", err)
```

pkg/types/encrypted_private_key.go
```diff
@@ -29,74 +31,81 @@ func (a *EncryptedPrivateKey) validate() error {
...
 }
 
-// PublicKey returns the public key corresponding to the private key.
+// PublicKey returns the public key corresponding to the private key. Password is destroyed.
 func (e *EncryptedPrivateKey) PublicKey(password *memguard.LockedBuffer, salt, nonce []byte) (*PublicKey, error) {
-	privateKeyInClear, err := PrivateKey(password, salt, nonce, e.Data)
+	privateKeyInClear, err := privateKey(password, salt, nonce, e.Data)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("PublicKey: %w", err)
```
gregLibert, line 108: This is also true here.

pkg/types/encrypted_private_key.go
```diff
@@ -112,7 +121,37 @@ func (e *EncryptedPrivateKey) PublicKey(password *memguard.LockedBuffer, salt, n
 	}, nil
 }
 
-func PrivateKey(password *memguard.LockedBuffer, salt, nonce, encryptedKey []byte) (*memguard.LockedBuffer, error) {
+// PrivateKeyTextInClear returns the private key in clear. Password is destroyed.
+func (e *EncryptedPrivateKey) PrivateKeyTextInClear(password *memguard.LockedBuffer, salt, nonce, encryptedKey []byte) (*memguard.LockedBuffer, error) {
+	privateKeyInClear, err := privateKey(password, salt, nonce, encryptedKey)
+	if err != nil {
+		return nil, fmt.Errorf("PrivateKeyTextInClear: %w", err)
```
gregLibert, line 128: Same here.

pkg/types/encrypted_private_key.go
```diff
@@ -112,7 +121,37 @@ func (e *EncryptedPrivateKey) PublicKey(password *memguard.LockedBuffer, salt, n
...
+	}
+
+	seed := ed25519.PrivateKey(privateKeyInClear.Bytes()).Seed()
+	privateKeyInClear.Destroy()
+
+	seedBuffer := memguard.NewBufferFromBytes(seed)
+
+	privateKey := e.Kind.Prefix() + base58.CheckEncode(seedBuffer.Bytes(), e.Version)
+	privateKeyBuffer := memguard.NewBufferFromBytes([]byte(privateKey))
```
gregLibert, line 137: Why seedBuffer is not destroyed?

pkg/types/encrypted_private_key.go
```diff
@@ -112,7 +121,37 @@ func (e *EncryptedPrivateKey) PublicKey(password *memguard.LockedBuffer, salt, n
...
+	return privateKeyBuffer, nil
+}
+
+// PasswordIsValid returns true if the password is valid for the account. It destroys the password.
+func (e *EncryptedPrivateKey) PasswordIsValid(password *memguard.LockedBuffer, salt, nonce, encryptedKey []byte) bool {
+	privateKeyInClear, err := privateKey(password, salt, nonce, e.Data)
+	if err != nil {
+		return false
+	}
+	privateKeyInClear.Destroy()
```
gregLibert, line 148: This line seems to be wronly formatted. Can you check the linter?

pkg/types/encrypted_private_key.go
```diff
@@ -112,7 +121,37 @@ func (e *EncryptedPrivateKey) PublicKey(password *memguard.LockedBuffer, salt, n
...
+
+	return err == nil
```
gregLibert, line 150: err is necessarly nil by construction here.

pkg/types/encrypted_private_key.go
```diff
@@ -112,7 +121,37 @@ func (e *EncryptedPrivateKey) PublicKey(password *memguard.LockedBuffer, salt, n
...
+
+	return privateKeyBuffer, nil
+}
+
+// PasswordIsValid returns true if the password is valid for the account. It destroys the password.
+func (e *EncryptedPrivateKey) PasswordIsValid(password *memguard.LockedBuffer, salt, nonce, encryptedKey []byte) bool {
+	privateKeyInClear, err := privateKey(password, salt, nonce, e.Data)
+	if err != nil {
+		return false
+	}
```
gregLibert, line 147: The logic of this function is weird. If the salt or the nonce is wrong, but the password is ok, you will also get an error. and the function PasswordIsValid will return false...
Furthermore, are you sure that except from an invalid salt, nonce or password, no other error are returned ?

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,246 @@
...
+)
+
+var (
+	ErrInvalidPrivateKey = errors.New("invalid private key")
```
gregLibert, line 23: Why the private key is having it's own error? What about reusing the InvalidParameter aggregation instead?

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,246 @@
...
+	ErrGeneratingKeyPair = errors.New("generating ed25519 key-pair")
```
gregLibert, line 24: This error is too specific and will make your code more complex to read in the error handling section.
I suggest to do the following:
```golang
var ErrRuntime = errors.New("runtime execution")
...
publicKeyBytes, privateKeyBytes, err := ed25519.GenerateKey(rand.Reader)
if err != nil {
	return nil, fmt.Errorf("key generation failed: %w: %w", ErrRuntime, err)
}
...
```
That way you'll be able to catch the ErrRuntime to return http code 500.

Note: the second %w is optional, you could use %v instead, but doing it you wrap the underlying error, exposing it to the caller if at somepoint he will need to differenciate different errors.

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,246 @@
...
+	ErrInvalidParameter  = errors.New("invalid parameter")
+)
+
+type Account struct {
+	Version      uint8                      `yaml:"Version"`
+	Nickname     string                     `yaml:"Nickname,omitempty"`
+	Address      *types.Address             `yaml:"Address,omitempty"`
```
gregLibert, line 31: Why do youo have omitempty here and not elsewhere ?

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,246 @@
...
+	version uint8,
+	nickname string,
+	address *types.Address,
+	salt [16]byte,
+	nonce [12]byte,
+	encryptedPrivateKey *types.EncryptedPrivateKey,
+	publicKey *types.PublicKey,
+) (*Account, error) {
+	if version > AccountLastVersion {
+		return nil, fmt.Errorf("%w: %d", ErrInvalidParameter, version)
```
gregLibert, line 48: The error message will be "invalid parameter: 0". Can you change it to something easier to understand ?

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	if !NicknameIsValid(nickname) {
+		return nil, fmt.Errorf("%w: %s", ErrInvalidNickname, nickname)
```
gregLibert, line 52: This is also true here and nickname is a parameter

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	if err := address.Validate(address.Version, object.UserAddress, object.SmartContractAddress); err != nil {
+		return nil, fmt.Errorf("validating address: %w", err)
```
gregLibert, line 56: You should use invalid parameter sentinel error

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	if err := encryptedPrivateKey.Validate(encryptedPrivateKey.Version, object.EncryptedPrivateKey); err != nil {
+		return nil, fmt.Errorf("validating encrypted private key: %w", err)
```
gregLibert, line 60: Ok, I guess you can review all errors

pkg/wallet/account/account.go
```diff
@@ -0,0 +1,246 @@
...
+	return a.CipheredData.PrivateKeyTextInClear(password, a.Salt[:], a.Nonce[:], a.CipheredData.Data)
+}
+
+// Sign signs the data with the private key and destroys the password.
+func (a *Account) Sign(password *memguard.LockedBuffer, data []byte) ([]byte, error) {
+	return a.CipheredData.Sign(password, a.Salt[:], a.Nonce[:], data)
+}
+
+func (a *Account) Marshal() ([]byte, error) {
+	yamlMarshaled, err := yaml.Marshal(a)
```
gregLibert, line 206: why don't you directly return yaml.Marshal(a) ? Except if you want to wrap it using  sentinel error, but you are not doing so...

gregLibert (changes requested):

pkg/wallet/account/account.go
Thykof: nickname and address are optional fields

gregLibert (approved):

PR #768, opened 2023-10-02
766 avoid unsafe message when installing on windows

thomas-senechal:


.github/workflows/build.yml
```diff
@@ -79,6 +79,16 @@ jobs:
           chmod +x build/bin/*/Contents/MacOS/*
           ditto -c -k build/bin/wallet-plugin.app wallet-plugin.app.zip
 
+      - name: Sign Windows binary
+        if: matrix.target == 'windows'
+        uses: ./.github/actions/sign-file
+        with:
+          file: $TARGET_NAME${{ matrix.ext }}
+          GCP_PEM_KEY: ${{ vars.GCP_PEM_KEY }}
+          GCP_KEYSTORE_ID: ${{ secrets.GCP_KEYSTORE_ID }}
```
Thykof, line 88: check if this way of passing secret is safe, because a the action side, the input are not view as secret from github action ci

gregLibert (approved):

Thykof (approved):

PR #780, opened 2023-10-09
Add From Nickname to sign

pivilartisant:
Screenshot 2023-10-09 at 15 34 31
callSC
transaction
executeSc

### DIFF
--- a/wails-frontend/src/pages/PasswordPromptHandler/PlainText/PlainText.tsx
+++ b/wails-frontend/src/pages/PasswordPromptHandler/PlainText/PlainText.tsx
@@ -7,7 +7,8 @@ import Intl from '@/i18n/i18n';
 import { maskAddress } from '@/utils';
 
 export function PlainText(props: PromptRequestData) {
-  const { PlainText, DisplayData, WalletAddress, Description } = props;
+  const { PlainText, DisplayData, WalletAddress, Description, Nickname } =
+    props;
 
   const toAddInHeigthDescription = Description ? 50 : 0;
   const toAddInHeigthDisplayData = DisplayData ? 50 : -50;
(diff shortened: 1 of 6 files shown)


wails-frontend/src/pages/PasswordPromptHandler/Sign.tsx
Thykof, line 1: 😻

Thykof (approved):

wails-frontend/src/pages/PasswordPromptHandler/PlainText/PlainText.tsx
```diff
@@ -19,33 +20,39 @@ export function PlainText(props: PromptRequestData) {
...
-        </div>
+        <>
+          <hr className="h-0.25 bg-neutral opacity-40 w-full" />
+
+          <div className="flex flex-col w-full h-fit">
+            <AccordionCategory
+              isChild={false}
+              iconOpen={<FiChevronDown />}
+              iconClose={<FiChevronUp />}
+              customClass={'!p-0'}
```
mazmassa, line 40: Do we need to force here!? It should not be used.

pivilartisant: Screenshot from 2023-10-10 10-36-51
for some odd reason yes

pivilartisant: Screenshot from 2023-10-10 10-39-06
the custom class is set correctly too

pivilartisant: but setting it to px-0 and py-0 seems to work correctly

mazmassa: Better! 
Force with `!` should be avoided. ;)

Tks for that!

mazmassa (approved):

PR #801, opened 2023-10-23
e2e test: create account

pivilartisant:
We are adding e2e browser create account  test coverage


web-frontend/cypress/e2e/acceptance/account/accountCreate.cy.tsx
```diff
@@ -0,0 +1,89 @@
...
+      compareSnapshot(cy, 'account-create-step-two');
+
+      cy.get('[data-testid="button"]').contains('Define a password').click();
+
+      cy.url().should('eq', `${baseUrl}/account-create-step-three`);
+      compareSnapshot(cy, 'account-create-step-three');
+
+      cy.get('[data-testid="button"]').contains('Skip').click();
+
+      server.createList('account', 1);
```
mazmassa, line 52: If is only one account you don't need use createList. Just `server.create('account');` should be fine!

web-frontend/cypress/e2e/acceptance/account/accountCreate.cy.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+      cy.url().should('eq', `${baseUrl}/account-create-step-one`);
+
+      cy.get('[data-testid="input-field"]').clear().type('testAccount');
+      cy.get('[data-testid="button"]').contains('Next').click();
+
+      cy.url().should('eq', `${baseUrl}/account-create-step-two`);
+
+      compareSnapshot(cy, 'wrong-account-name-format');
+    });
```
mazmassa, line 87: I am not sure about this one. 🤔 

We have one test that assert when all goes fine: ✅ 
We have one test that assert when empty field: ✅ 

What would this one add more?

pivilartisant: true, I wanted to cover as closely as possible user behavior but if it's not necessary I will delete !

mazmassa: >closely as possible user behavior
You did! Both cases. When fails, when not! :)

mazmassa (approved):

Thykof (approved):
LGTM

PR #804, opened 2023-10-24
prepush hook don't run cypress tests

Thykof:
@pivilartisant @mazmassa I propose to not run ent2end cypress test in the git pre push hook. Let's keep it only in the CI, WDYT?


pivilartisant:
> @pivilartisant @mazmassa I propose to not run ent2end cypress test in the git pre push hook. Let's keep it only in the CI, WDYT?

I have no opposition to this

pivilartisant (approved):

mazmassa (approved):

mazmassa:
> @pivilartisant @mazmassa I propose to not run ent2end cypress test in the git pre push hook. Let's keep it only in the CI, WDYT?

Sure!
I would say to not run any, except if we want. :)

PR #818, opened 2023-10-26
817 hide expiry period and max gas in the wails pop ups of sc type operations

Thykof:


thomas-senechal (approved):
I guess it would be better to put those kind of information in a "details" section that is collapsable, so by default, its hidden, but if needed, users can access it

PR #821, opened 2023-10-27
819-1: operation cost: wails front

Thykof:
This PR improve the sign prompt and allow the user to modify the fees, but the modification is not taken into account (see following PR branch 819-2)


wails-frontend/src/pages/PasswordPromptHandler/OperationCost.tsx
```diff
@@ -0,0 +1,170 @@
...
+                    disabled
+                    value={formatStandard(coins)}
+                  />
+                </div>
+              )}
+              <div className="flex flex-row justify-end gap-1">
+                {isEditing ? (
+                  <>
+                    <button
+                      className="flex flex-row hover:cursor-pointer"
```
pivilartisant, line 139: tailwind tip: when using the tag "flex" attr "flex-row" is default, so you can safely remove  "flex-row"

wails-frontend/src/pages/PasswordPromptHandler/Sign.tsx
```diff
@@ -105,14 +119,47 @@ export function Sign() {
           {(() => {
             switch (signData.OperationType) {
               case OPER_CALL_SC:
-                return <CallSc {...signData} />;
+                return (
```
pivilartisant, line 122: Concerning the fees being disabled during, i believe that passing isEditing as a param to OperationCost would make sense... tdb in small follow up pr

pivilartisant (approved):
I left 3 comments, nothing major but otherwise: LGTM!

wails-frontend/src/pages/PasswordPromptHandler/OperationCost.tsx
Thykof: ok but there are many other flex flex-row in the code, maybe another PR to simplify all this

pivilartisant: Totally

thomas-senechal (approved):
Backend looks good to me

PR #826, opened 2023-11-06
Change Currency for Money

pivilartisant:


web-frontend/cypress/downloads/downloads.html
```diff
@@ -0,0 +1 @@
+Cr24
```
Thykof, line 1: do we need this?

pivilartisant: no !

Thykof (approved):

PR #832, opened 2023-11-07
819 3 retrocompatibility allow edit fees

Thykof:


pivilartisant (approved):
LGTM

peterjah (approved):
Personnaly i would prefer not merge this and keep the breaking change.
I see no use case where dapp would forbid the fee edition.

thomas-senechal (approved):

PR #847, opened 2023-11-14
cache a copy of the cipheredPrivateKey

Thykof:


internal/handler/wallet/sign.go
```diff
@@ -360,21 +363,28 @@ func (w *walletSign) privateKeyFromCache(
...
-	if err != nil {
-		return nil, fmt.Errorf("%w: %w", utils.ErrCache, err)
+	byteValue, ok := value.([]byte)
+	if !ok {
+		return nil, fmt.Errorf("%w: %s", utils.ErrCache, "value is not a byte array")
 	}
 
-	cipheredPrivateKey := memguard.NewBufferFromBytes(buf.Bytes())
+	cacheValue := make([]byte, ed25519.PrivateKeySize)
+	copy(cacheValue, byteValue)
```
peterjah, line 376: are you sure this copy is needed ? i don't understand why

Thykof: ho, let me try without

gregLibert (approved):

internal/handler/wallet/sign.go
gregLibert: > are you sure this copy is needed ? i don't understand why

+1

Thykof: it's needed, I try without and the second time we retrieve the ciphered private key from the cache, the unciphered private key is not the same, so the signature is invalid.

Thykof: before we did: 

```go
buf := new(bytes.Buffer)
err = binary.Write(buf, binary.LittleEndian, value)
```

so now
```
cacheValue := make([]byte, ed25519.PrivateKeySize)
copy(cacheValue, byteValue)
```

is equivalent but better because we specify the length of the byte array and it's more readable than `binary.Write(buf, binary.LittleEndian, value)`

PR #849, opened 2023-11-17
avoid invalid account duplicate

Thykof:


Thykof:
close in favor of another solution to the same bug: https://github.com/massalabs/station-massa-wallet/pull/850

PR #856, opened 2023-12-18
Add chain

Thykof:


thomas-senechal (approved):

damip:
CI seems to fail

Thykof:
> CI seems to fail

it fails because buildnet is not handling chain id yet.

peterjah (approved):

PR #869, opened 2024-01-03
fix: update account: handle duplicate nickname error

Thykof:


internal/handler/wallet/update.go
```diff
@@ -36,7 +37,7 @@ func (w *walletUpdateAccount) Handle(params operations.UpdateAccountParams) midd
 
 	newAcc, err := w.handleUpdateAccount(acc, params.Body.Nickname)
 	if err != nil {
-		return newErrorResponse(err.Error(), "", http.StatusInternalServerError)
+		return newErrorResponse(err.Error(), utils.ErrDuplicateNickname, http.StatusBadRequest)
```
ASAPSegfault, line 40: A 409 may be more appropriate

ASAPSegfault (approved):

PR #879, opened 2024-01-04
web: set onClickLogo

Thykof:
todo:
- [x] update ui kit
- [x] set the callback


thomas-senechal (approved):
Not tested but looks good to me

PR #890, opened 2024-01-24
Fix regressions

Thykof:


thomas-senechal (approved):
Nice fix ! 
Works great 

Just tests to fix :/

PR #904, opened 2024-04-04
Fix asset decimal pointer

Thykof:


Thykof:
@gregLibert chatgpt4 gave me the answer but I don't fully understand. Could you explain why we had the same pointer for decimals value before this fix?

pkg/assets/asset.go
```diff
@@ -85,11 +85,12 @@ func (s *AssetsStore) loadaccountsStore(assetsJSONPath string) error {
 		}
 
 		for _, asset := range accountData.Assets {
+			decimals := asset.Decimals // Copy the decimals value to avoid a pointer to a local variable
 			assetInfo := models.AssetInfo{
 				Address:  asset.ContractAddress,
 				Name:     asset.Name,
 				Symbol:   asset.Symbol,
-				Decimals: &asset.Decimals,
+				Decimals: &decimals,
```
bilboquet, line 93: are you sure you want to keep the `&`?

Thykof: yes because I need a pointer

Thykof:
before merging, check: we have a similar loop in the save function, maybe we need also to copy the decimal value

pkg/assets/asset.go
bilboquet: so I found your comment above 
> // Copy the decimals value to avoid a pointer to a local variable 

a bit misleading because it seems that you're exactly getting a pointer to a local variable. But maybe there's a Go trick, I'd like to know, here.

PR #909, opened 2024-04-10
web: minimal fees and fix round amount

Thykof:
# content
- [x] fix round amount in the frontend
- [x] display all amounts in MAS, not in nano MAS
- [x] check minimal fees in wails prompt: minimal fees is the minimal fees returned by the get_status of the node we are connected to
- [x] default fees in wails prompt: fee value given in the call of minimal fees of the node if bellow
- [x] check maximum fees in wails prompt: maximal fees is the MAS total supply: 1,000,000,000 MAS
- [x] check advanced fees preset: 0.01, 0.03 and 0.1 (can be argue)

## QA minimal fees:
image
image

## QA the fix of round amount:
To QA fees to low you need to be connected to a node with minimal fees feature. You can do that by running a local sandbox on main branch of massa repo, and add the config in the network setting of massa station:
```
sandbox:
  URLs:
  - http://localhost:33035
  ChainID: 77
```

account selection:
image

home page
image

receive: generate link
image

send:
image

contact list:
image

confirmation:
image

asset:
image


thomas-senechal (approved):
Not tested but the code looks good to me

pivilartisant (approved):

SlnPons:
QA testing output
**1. Sending / Receiving flow:** 
* Advance settings: the custom fees field always displays `0.01`. The field should be empty by default.
* Confirmation page: inconsistency between the decimals shown for the amount (always 9) and the fees (what the user typed). It should be consistent and the same behaviour should be applied across all the app. ---> I suggest to ignore this in this PR and solve this problem across all the app in a next mission.
* Sign pop-up: no fee as shown even if I have some. 
image

**2. Settings page:**
* The back up button is off: it doesn't work when you click on it. It should open the back-up pop-up.
* Delete account: same as above

PR #923, opened 2024-04-29
Wail ft transfer

Thykof:
first PR for the transfer FT feature


thomas-senechal (approved):
The code looks good to me
I didn't tested it tho

PR #926, opened 2024-04-29
try fix ci

Thykof:
The build CI fail since https://github.com/massalabs/station-massa-wallet/commit/e5b47f77b0de0b0233053d647536128a2dbd512e

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -77,8 +77,7 @@ jobs:
         if: matrix.target == 'darwin'
         shell: bash
         run: |
-          # install wails with amd64 arch, even for arm64, to be able to use it for the build
-          GOARCH=amd64 go install github.com/wailsapp/wails/v2/cmd/wails@v2.6.0
+          export PATH=$PATH:~/go/bin/darwin_amd64
           task build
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -77,8 +77,7 @@ jobs:
         if: matrix.target == 'darwin'
         shell: bash
         run: |
-          # install wails with amd64 arch, even for arm64, to be able to use it for the build
-          GOARCH=amd64 go install github.com/wailsapp/wails/v2/cmd/wails@v2.6.0
+          export PATH=$PATH:~/go/bin/darwin_amd64
```
thomas-senechal, line 80: We remove the installation of wails ? 🤔

Thykof: it's already installed in task install, in action install.yaml

thomas-senechal (approved):

.github/workflows/build.yml
Thykof, line 80: runner macos-latest is now macos 14 with M1 CPU, so arm64 arch. So to fix the build on amd, we add the go/bin/darwin_amd64 directory in the path

PR #929, opened 2024-05-02
Disable switch account in confirmation

Thykof:


thomas-senechal (approved):
The code looks good to me
But I didn't tested it

PR #935, opened 2024-05-07
Add networkStore

pivilartisant:


pivilartisant:
@Thykof I think we should add client, isMainnet and nickname (if maybe not necessary) to the store... I didn't want to add this the the AppStore... let me know if you agree with this and the name

Thykof:
Hi @pivilartisant can you provide context for this PR?

pivilartisant:
> Hi @pivilartisant can you provide context for this PR?

Hi yes, it aims to standardize the use of stores across all our products.

PR #947, opened 2024-05-22
Home page ui fixes

pivilartisant:
We are fixing new mns feature to add QA feedback:
- down case D in "Domains"
- Truncate part of the dns list for better ux (user must understand they can scroll)
- Align right address and mns in confirmation page 
- Reduce tooltip size

### DIFF
--- a/api/server/models/account.go
+++ b/api/server/models/account.go
@@ -41,7 +41,7 @@ type Account struct {
 	Nickname Nickname `json:"nickname"`
 
 	// status
-	// Enum: [ok corrupted]
+	// Enum: ["ok","corrupted"]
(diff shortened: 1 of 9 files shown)


Thykof (commented):
React Hook useEffect has a missing dependency: 'resetDomainList'. Either include it or remove the dependency array

you hard code the px size, are you sure it's a good idea? I know the app is not responsive...

pivilartisant:
> React Hook useEffect has a missing dependency: 'resetDomainList'. Either include it or remove the dependency array

Adding this dependency causes infinite re-renders i will try to find another way

Thykof (approved):

PR #952, opened 2024-06-05
Asset doc and perf

Thykof:
need the update of the ui kit after merge of https://github.com/massalabs/ui-kit/pull/455


peterjah:
It seems the same usd price is fetched several times (ETHUSD and USD tickers) moreover fetching USD price make no sense to me. why not just return 1$ ??

peterjah (approved):

thomas-senechal (approved):

pivilartisant (approved):

PR #953, opened 2024-06-05
Purrfect Universe

AltayDev:
For the icon : 

https://www.purrfectuniverse.com/charlie.webp


Thykof (approved):
great, if you have a PUR token on the buildnet you can add it also

Thykof:
For the icon we need to add in the https://github.com/massalabs/Ui-Kit, @SlnPons @pivilartisant

PR #954, opened 2024-06-05
add chain id in asset

Thykof:


internal/handler/wallet/get_all_assets.go
```diff
@@ -130,13 +135,11 @@ func (g *getAllAssets) getAssetsData(acc *account.Account) ([]*assets.AssetInfoW
 		go func(asset *assets.AssetInfoWithBalances) {
 			defer wg.Done()
 
-			// First, check if the asset exists in the network
-			if !g.massaClient.AssetExistInNetwork(asset.AssetInfo.Address) {
```
peterjah, line 134: maybe this check is still relevant ?

peterjah (approved):

internal/handler/wallet/get_all_assets.go
Thykof: no because before that we filter by chain id

PR #955, opened 2024-06-05
Doc add token

Thykof:


internal/handler/wallet/get_all_assets_test.go
```diff
@@ -23,8 +23,6 @@ func TestGetAllAssetsHandler(t *testing.T) {
 	assetsWithBalance := getAssets(t, api, nickname)
 
 	// Assert that assetsWithBalance contains the expected data
-	assert.Len(t, assetsWithBalance, 8, "the assets list should have 8 items")
-
```
Thykof, line 27: When a builder add a token he needs to update the unit test here. I think this test is not really relevant.

peterjah: Why not just get the nb of asset dynamically instead of hardcode it

Thykof: I need a refacto of another branch to do that, let's merge this if it's ok

peterjah (approved):

peterjah (approved):

PR #961, opened 2024-06-11
BSC tokens and update ui kit format amount

Thykof:
- [x] merge https://github.com/massalabs/ui-kit/pull/459 and update ui kit 
- [x] add BSC tokens addresses


Thykof:
go unit tests should be solved when we specify the correct mainnet addresses of the 2 tokens

pivilartisant (approved):

PR #978, opened 2024-12-17
add global config file manager

peterjah:


thomas-senechal (approved):
Looks good to me

fleandrei:
To fix this Test web-frontend issue, I added ".js" at the end of `import getCompareSnapshotsPlugin from 'cypress-image-diff-js/dist/plugin'; ` in web-frontend/cypress.config.ts

PR #989, opened 2025-02-04
fix wails operationcost show deployement info (ex bytecode storage cost ...) when not deploying

fleandrei:


wails-frontend/src/pages/PasswordPromptHandler/SignComponentUtils/OperationCost.tsx
```diff
@@ -43,9 +44,18 @@ export interface OperationCostProps {
 }
 
 export function OperationCost(props: OperationCostProps) {
-  const hideCoins = props.coins === undefined;
+  const hideCoins = props.coins ? props.coins === '0' : true;
```
peterjah, line 47: `const hideCoins = !props?.coins`

wails-frontend/src/pages/PasswordPromptHandler/SignComponentUtils/OperationCost.tsx
```diff
@@ -43,9 +44,18 @@ export interface OperationCostProps {
...
   const hideByteCodeCost = props.DeployedByteCodeSize === 0;
-  const hideDeployedCoins = props.DeployedCoins === undefined;
+  const hideDeployedCoins = props.DeployedCoins
```
peterjah, line 49: const hideDeployedCoins = !props?.DeployedCoins

peterjah (approved):

wails-frontend/src/pages/PasswordPromptHandler/SignComponentUtils/OperationCost.tsx
fleandrei: props cannot be null or undefined so I don't think it is useful to use a '?'.
If props.coins is '0', `!props?.coins` would be false. So hideCoins would be false.

wails-frontend/src/pages/PasswordPromptHandler/SignComponentUtils/OperationCost.tsx
fleandrei: If props.DeployedCoins is '0', `!props?.DeployedCoins` would be false and hence hideDeployedCoins would be false.

PR #994, opened 2025-02-14
add All contract in signrule modal

peterjah:


web-frontend/src/i18n/en_US.json
```diff
@@ -112,6 +112,7 @@
       "disable-password-prompt-tooltip": "This will disable the password prompt for operations coming from the linked contract address, but will still require your confirmation.",
       "errors": {
         "add": "An error occurred while adding the sign rule",
+        "alreadyExist": "A similar rule already exists for account",
```
thomas-senechal, line 115: details but I would put something like:
```suggestion
        "already-exist": "A similar rule already exists for account",
```
to keep the format

PR #997, opened 2025-03-03
Fix qa feedback

thomas-senechal:


web-frontend/src/pages/Settings/SettingsSignRule.tsx
```diff
@@ -201,7 +204,7 @@ function SignRuleListItem(props: SignRuleListItemProps) {
             triggerClassName="truncate w-full"
             tooltipClassName="mas-caption"
           >
-            {rule.name}
+            {rule.name.length === 0 ? '-' : rule.name}
```
fleandrei, line 207: I think checking rule.name.length is not useful because there is already a check line 200 : rule.name && rule.name.length > 0

thomas-senechal: houla
you are right, this shouldn't be like that 
will fix that

nice catch !

fleandrei (approved):

